PR #464, opened 2013-09-17
Fix compound arrays

alno:
Calling type_cast for array elements, which fixes serialization of int arrays from strings and compound arrays (for example, arrays of json objects).

Also fixes deserializing compound arrays (calling type_cast for deserialized element)


kares:
this issue should now be irrelevant as its handled by the AR-type system introduced in 4.2

PR #484, opened 2013-10-22
Fall back to rollback_savepoint for MS SQL Server driver

seandmccarthy:
For nested transactions, ActiveRecord makes use of savepoints. When committing these transactions, ActiveRecord will call release_savepoint.

See https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb#L199

It seems that both the jTDS and Microsoft's own JDBC driver report that they support save points, however the latest Microsoft SQL Server driver (http://msdn.microsoft.com/en-us/sqlserver/aa937724.aspx) does not implement release_savepoint:

```
jruby-1.7.5 :003 > ActiveRecord::Base.connection.jdbc_connection.release_savepoint('foo')
Java::ComMicrosoftSqlserverJdbc::SQLServerException: This operation is not supported.
```

[This article](http://technet.microsoft.com/en-us/library/ms378414.aspx]) seems to show how Microsoft expect savepoints to be used with their driver. They use rollback with the savepoint name instead of invoking the JDBC releaseSavepoint method.

This pull request adds release_savepoint to the MSSQL adapter, overriding the version in the JdbcAdapter base class. It checks to see if the MSSQL adapter is using the Microsoft driver and if so calls rollback_savepoint instead.

### DIFF
--- a/lib/arjdbc/mssql/adapter.rb
+++ b/lib/arjdbc/mssql/adapter.rb
@@ -6,13 +6,14 @@ require 'arjdbc/mssql/limit_helpers'
 require 'arjdbc/mssql/lock_methods'
 require 'arjdbc/mssql/column'
 require 'arjdbc/mssql/explain_support'
+require 'arjdbc/mssql/driver_info'
 
 module ArJdbc
   module MSSQL
     include LimitHelpers
     include Utils
-
     include ExplainSupport
+    include DriverInfo
 
     # @private
     def self.extended(adapter)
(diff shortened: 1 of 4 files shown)


jjustaments:
Hi,

We try to use MS JDBC driver but our application rspec tests failed.
The reason is this "temporary" hack following line https://github.com/jruby/activerecord-jdbc-adapter/blob/54592892df12ca1b1b1a5565fe67bd0ca2d5e60f/lib/arjdbc/mssql/adapter.rb#L719 rollback_savepoint can't be used instead of release_savepoint because these are totally different things. I understand that MS SQL Server doesn't have release savepoint feature but we cannot just rollback database changes in this case. For testing purpose, we commented out rollback_savepoint line and our tests start working.

PR #503, opened 2013-11-25
[DB2] Add activerecord-jdbcas400-adapter

pierrickrouxel:
Add an AS/400 adapter. 

I hope see it in RubyGem in 1.3.4 :)

Thanks !


aaronbartell:
For historical purposes you can find the new repository [here](https://github.com/pierrickrouxel/activerecord-jdbcas400-adapter)

PR #580, opened 2014-08-29
fix error so that jruby add_column will respect "after" option

zben:
In activerecord, you can specify add_column with "after" option to tell what column to append the new column after. In jRuby, this option is ignored. This patch should be able to fix it. I am not sure the existing class was tested at all. I will be happy to pair with someone to pair on writing the test.


kares:
not going in as it seems no longer relevant [since SchemaCreation was backported](https://github.com/jruby/activerecord-jdbc-adapter/commit/dba1f952fb8c7058677137182c0b0b1895afe54f) (in 1.3.14), the confusion likely came due AR 4.0 having 2 `add_column_options!` [methods](https://github.com/rails/rails/blob/4-0-stable/activerecord/lib/active_record/connection_adapters/abstract_mysql_adapter.rb#L9-L20), the PR simply changes the wrong one.

PR #608, opened 2014-12-08
fixes #590

### LINKED ISSUE #590: Issue when using joins with distinct
We have a basic user to roles mapping with a roles_users table in between.

`User.joins(:roles_users).distinct.where(id: 1).first`

would cause the error

`ActiveRecord::JDBCError: The multi-part identifier "users.id" could not be bound`

while 

`User.joins(:roles_users).where(id: 1).first`

works fine.

We are using an MSSQL Database, therefore we use the `activerecord-jdbcmssql-adapter` (newest version). Here are the two sql statements that are being generated:

_cruby_:
```
EXEC sp_executesql N'SELECT DISTINCT TOP (1) [users].* 

FROM [users] INNER JOIN [roles_users] ON [roles_users].[user_id] = [users].[id] AND [roles_users].[deleted_at] IS NULL WHERE [users].[first_name] = N''abc'

' ORDER BY [users].[id] ASC'
```
_jruby_:
```
SELECT t.* FROM ( SELECT ROW_NUMBER() OVER(ORDER BY [users].[id] ASC) AS _row_num, t.* FROM (SELECT  DISTINCT [users].* 

FROM [users] INNER JOIN [roles_users] ON [roles_users].[user_id] = [users].[id] AND [roles_users].[deleted_at] IS NULL WHERE [users].[first_name] = N'abc'

) AS t ) AS t WHERE t._row_num BETWEEN 1 AND 1
```

We have tried it with different jruby versions: `jruby-1.7.9` and `jruby-1.7.13` on `Rails 4.1.1` and `Rails 4.1.6`.

====
you can support MS-SQL fixes at [BountySource](https://www.bountysource.com/issues/5135362-issue-when-using-joins-with-distinct)

falti:
Run the test

``` 
$ rake test_mssql TEST=test/db/mssql/unit_test.rb

```

I could not run this fix against MS SQL database yet as I have some trouble with running this from a Mac.

### DIFF
--- a/lib/arjdbc/mssql/limit_helpers.rb
+++ b/lib/arjdbc/mssql/limit_helpers.rb
@@ -39,7 +39,6 @@ module ArJdbc
           end
 
           if distinct # select =~ /DISTINCT/i
-            # order = order.gsub(/([a-z0-9_])+\./, 't.')
             order = order.gsub(/(\[[a-z0-9_]+\]|[a-z0-9_]+)\./, 't.')
             new_sql = "SELECT t.* FROM "
             new_sql << "( SELECT ROW_NUMBER() OVER(#{order}) AS _row_num, t.* FROM (#{select} #{rest_of_query}) AS t ) AS t"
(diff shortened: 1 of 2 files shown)


kares:
commits are picked on **1-3-stable** and are expected to land in **1.3.18** ... thanks again for the PR

Issue #611, opened 2015-01-08
ArgumentError: wrong number of arguments in sqlite adapter


michaelachrisco:
When using the schemaplus gem, the following error occurs:
``` Bash
$ rails s
ArgumentError: wrong number of arguments calling `initialize` (5 for 4)
                              initialize at /Users/michaelchrisco/.rvm/gems/jruby-1.7.16.1/gems/activerecord-jdbc-adapter-1.3.13/lib/arjdbc/sqlite3/adapter.rb:572
                                 columns at /Users/michaelchrisco/.rvm/gems/jruby-1.7.16.1/gems/activerecord-jdbc-adapter-1.3.13/lib/arjdbc/sqlite3/adapter.rb:361
                                     map at org/jruby/RubyArray.java:2412
                                 columns at /Users/michaelchrisco/.rvm/gems/jruby-1.7.16.1/gems/activerecord-jdbc-adapter-1.3.13/lib/arjdbc/sqlite3/adapter.rb:355
                    prepare_default_proc at /Users/michaelchrisco/.rvm/gems/jruby-1.7.16.1/gems/activerecord-4.1.5/lib/active_record/connection_adapters/schema_cache.rb:93
                                    call at org/jruby/RubyProc.java:271
                                 default at org/jruby/RubyHash.java:708
                                      [] at org/jruby/RubyHash.java:1092
                                 columns at /Users/michaelchrisco/.rvm/gems/jruby-1.7.16.1/gems/activerecord-4.1.5/lib/active_record/connection_adapters/schema_cache.rb:44
                                 columns at /Users/michaelchrisco/.rvm/gems/jruby-1.7.16.1/gems/activerecord-4.1.5/lib/active_record/model_schema.rb:214
                columns_with_schema_plus at /Users/michaelchrisco/.rvm/gems/jruby-1.7.16.1/gems/schema_plus-1.8.2/lib/schema_plus/active_record/base.rb:23
                            columns_hash at /Users/michaelchrisco/.rvm/gems/jruby-1.7.16.1/gems/activerecord-4.1.5/lib/active_record/model_schema.rb:223
               subclass_from_attributes? at /Users/michaelchrisco/.rvm/gems/jruby-1.7.16.1/gems/activerecord-4.1.5/lib/active_record/inheritance.rb:180
[...]

PR #612, opened 2015-01-08
[sqlite3] fix issue with initialize

michaelachrisco:
Fixes: #611


kares:
thanks, but this seems to change the semantics of `initialize` - we need to keep it compatible with several AR versions as well as our own usage ... would be good to have a failing piece of column code at #611

michaelachrisco:
Yes I am seeing this change failing other tests on Travis. Ill give it another go.

Issue #623, opened 2015-02-21
Avoid use of javax.naming package on Android


donv:
Hi!

When running with master on Android (Ruboto), I get the following error:

```
20150221 102153.947934 W/dalvikvm( 9190): VFY: unable to find class referenced in signature (Ljavax/naming/InitialContext;)
20150221 102153.958084 I/dalvikvm( 9190): Could not find method javax.naming.InitialContext.lookup, referenced from method arjdbc.jdbc.RubyJdbcConnection.jndi_lookup
20150221 102153.958119 W/dalvikvm( 9190): VFY: unable to resolve virtual method 8608: Ljavax/naming/InitialContext;.lookup (Ljava/lang/String;)Ljava/lang/Object;
20150221 102153.958133 D/dalvikvm( 9190): VFY: replacing opcode 0x6e at 0x0008
20150221 102153.958144 W/dalvikvm( 9190): VFY: unable to resolve exception class 1165 (Ljavax/naming/NamingException;)
20150221 102153.958158 W/dalvikvm( 9190): VFY: unable to find exception handler at addr 0x13
20150221 102153.958241 W/dalvikvm( 9190): VFY:  rejected Larjdbc/jdbc/RubyJdbcConnection;.jndi_lookup (Lorg/jruby/runtime/ThreadContext;Lorg/jruby/runtime/builtin/IRubyObject;Lorg/jruby/runtime/builtin/IRubyObject;)Lorg/jruby/runtime/builtin/IRubyObject;
20150221 102153.958265 W/dalvikvm( 9190): VFY:  rejecting opcode 0x0d at 0x0013
20150221 102153.958284 W/dalvikvm( 9190): VFY:  rejected Larjdbc/jdbc/RubyJdbcConnection;.jndi_lookup (Lorg/jruby/runtime/ThreadContext;Lorg/jruby/runtime/builtin/IRubyObject;Lorg/jruby/runtime/builtin/IRubyObject;)Lorg/jruby/runtime/builtin/IRubyObject;
20150221 102153.961599 W/dalvikvm( 9190): Verifier rejected class Larjdbc/jdbc/RubyJdbcConnection;
```

Can we avoid loading the ```javax.naming``` package on Android when using the sqlite3 adapter?

donv:
The changes from @kares have fixed this.

PR #624, opened 2015-02-21
Avoid direct reference to javax.naming for Android

donv:
Hi @kares !

Here is a workaround for issue #623 .  Could you see if it is OK to merge?  We loose compile time checking of other checked exceptions (currently none), but it should be caught and reported at runtime.


kares:
@donv that looks fine - maybe I would add a comment so there's a less of a chance of going "back"

kares:
did smt end in the end, since NamingException has sub-classes ... but it all should achieve the same for you

PR #627, opened 2015-02-26
Fix InformixAdapter for Rails 4.0/4.1

### LINKED ISSUE #622: Informix adapter undefined method 'column_types' with Rails 4.x
```
> Image.count
   (97.0ms)  SELECT COUNT(*) FROM images 
NoMethodError: undefined method `column_types' for [{"(count(*))"=>#><BigDecimal:5eb2172,'114234.0',7(8)>}]:Array
from /usr/local/Cellar/jruby/1.7.19/libexec/lib/ruby/gems/shared/gems/activerecord-4.1.9/lib/active_record/relation/calculations.rb:268:in `execute_simple_calculation'
```

This looks like whatever was done to fix #262 was not done for Informix.
(Rails 3.x works as expected.)

shado23:
Fixes #622.

This doesn’t add Rails 4.2 support.

### DIFF
--- a/lib/arjdbc/informix/adapter.rb
+++ b/lib/arjdbc/informix/adapter.rb
@@ -1,4 +1,4 @@
-require 'arjdbc/jdbc/serialized_attributes_helper'
+require 'arjdbc/util/serialized_attributes'
 
 module ArJdbc
   module Informix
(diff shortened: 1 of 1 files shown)


kares:
thanks @shado23 ... will put this on to 1-3-stable (first) and it will be in the next 1.3.x release

shado23:
Thanks.

I just noticed that I had accidentally removed the JdbcConnection alias, so I've re-added it and re-pushed.

kares:
great, did you manage to run tests `rake test_informix` ? just interested, there's [some](https://github.com/jruby/activerecord-jdbc-adapter/blob/master/test/db/informix/simple_test.rb) but you can simply [pull in](https://github.com/jruby/activerecord-jdbc-adapter/blob/master/test/db/firebird/simple_test.rb#L4-L8) [more](https://github.com/jruby/activerecord-jdbc-adapter/blob/master/test/db/oracle/reset_column_information_test.rb) to verify it's more or less compatible

shado23:
Unfortunately, I didn't manage to run the tests. I don't have a local Informix installation and I didn't have the time to see what would be necessary to get it to run against an external server (without blowing away existing databases). I'll see if I can make that happen.

kares:
on **1-3-stable** as https://github.com/jruby/activerecord-jdbc-adapter/commit/0a333d49b6c930e16944e5dcbf1fca1040baa618 (part of **1.3.15** release)

PR #631, opened 2015-03-15
Fixed issue #692 (for 4.2 create table with reference failed, created index twice) 

akarazniewicz:
Hi

Please review fix for #692. I know - this is probably bigger problem with AR 4.2 compatibility, hope it may help somehow however. I'm working on few more 4.2 issues and may pull some additional fixes. Added necessary test case.

### DIFF
--- a/lib/arjdbc/mysql/adapter.rb
+++ b/lib/arjdbc/mysql/adapter.rb
@@ -224,6 +224,11 @@ module ArJdbc
     end
 
     # @override
+    def supports_indexes_in_create?
+      true
+    end
+
+    # @override
     def supports_transaction_isolation?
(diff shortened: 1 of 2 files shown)


kares:
looking good, thanks ... maybe this would be needed for a couple of other adapters as well.

akarazniewicz:
Doh. Test case fails for AR<4.2 as add_foreign_key is new in 4.2 migrations (https://github.com/rails/rails/blob/7ba24831d7f809d54e88607db8f7d21780efbdfb/guides/source/4_2_release_notes.md#foreign-key-support). Fixed.

kares:
on **1-3-stable** now ... thanks!

PR #637, opened 2015-04-08
fix quote method for Hashes and Jsonb columns in postgres adapter

jenniferalons:


kares:
cherry-picked as https://github.com/jruby/activerecord-jdbc-adapter/commit/969e6ffff6a1a9d48536de4fd74bdc9d6f84fb92 ... thanks!

PR #639, opened 2015-04-13
Change jdbc-h2 dependency version to 1.4

nubbel:
Hi, we want to give H2 1.4 a shot in order to experiment with the new file locking method FS.
Unfortunately the current version of activerecord-jdbch2-adapter enforces jdbc-h2 version 1.3, which hinders us to do so.
This PR changes the version dependency to 1.4 (>= 1.3 would also be fine for us).


nubbel:
:+1:

PR #642, opened 2015-04-20
Update H2 driver to 1.4.187

nubbel:


nubbel:
The CI build passes now, can you merge please?

kares:
done, thanks ... guess we need to release this as well.

for the future, we do not have to gem release all driver betas coming out (just those with fixes or worth the new features) ... we should generally wait for final.

nubbel:
Thanks for merging!
I wanted to have the latest driver because we're experiencing issues with CLOBs currently. The 1.4.187 release notes contains lots of mentions of LOBs and the like. 
In the future I will just reference my branch and won't bug you anymore until 1.4 is final.

kares:
it's not a problem when there are issues ... just mention them that you've been experiencing issues with latest and the new beta resolved them. I only released 185 for the same reason as 181 had some issues.

1.4.187 is at rubygems, thanks

Issue #532, opened 2014-02-13
1.3.6 Breaks ordering on an aggregate in MSSQL


sanelson2000:
I previously added this as a comment to https://github.com/jruby/activerecord-jdbc-adapter/pull/529, but I think maybe I should have just created a new bug for it, so that's what I'm doing.

This patch appears to have broken something to do with ordering on an aggregate.

I have the following code:
```ruby
ConsensusWorkspace
        .select('w.*, count(o.object_id) num_objects')
        .joins("w inner join #{ConsensusObject.table_name} o on o.workspace_id = w.workspace_id" )
        .where('w.workspace_id > 0 and w.workspace_id IN (?)', accessible_workspace_ids)
        .group('w.workspace_id, w.workspace, w.description, w.person_id, w.status, w.verification_required, w.approval_process_id, w.team_id, w.last_modified_dt')
        .order('count(o.object_id) DESC')
        .limit(@workspace_count)
```

In 1.3.5, this generates the following valid query for MSSQL (it also works for Postgres, Oracle, and SQLite):
```sql
SELECT t.*
FROM (
	SELECT ROW_NUMBER() OVER (
			ORDER BY count(o.object_id) DESC
			) AS _row_num
		,w.*
		,count(o.object_id) num_objects
	FROM [tp_workspace] w
	INNER JOIN tp_object o ON o.workspace_id = w.workspace_id
	WHERE (
			w.workspace_id > 0
			AND w.workspace_id IN (N'0', N'8', N'9', N'11', N'12', N'13', N'15', N'16', N'18', N'21', N'22', N'23', N'24', N'25', N'26', N'28', N'29', N'30', N'31', N'32', N'33', N'40', N'41', N'42', N'43', N'44', N'45', N'52', N'53', N'54', N'55', N'56', N'57', N'62', N'63', N'64', N'65', N'66', N'67', N'58', N'59', N'60', N'61', N'68', N'69', N'46', N'47', N'48', N'49', N'50', N'51', N'34', N'35', N'36', N'37', N'38', N'39', N'20', N'77', N'78', N'79', N'80', N'81', N'82', N'84', N'86')
			)
	GROUP BY w.workspace_id
		,w.workspace
		,w.description
		,w.person_id
		,w.STATUS
		,w.verification_required
		,w.approval_process_id
		,w.team_id
		,w.last_modified_dt
	) AS t
WHERE t._row_num BETWEEN 1
		AND 10
```

[...]

PR #646, opened 2015-05-06
Fix broken limit_helpers

### LINKED ISSUE #645: Broken ordering on an aggregate in MSSQL
We see the same wrong behavior as described in #532. 

When using `limit` and `offset` the order-part gets wrapped in `MIN()` which breaks the query.

Versions we tried that fail: 1.3.16, 1.3.14 and 1.3.6
Last working version: 1.3.5

fpauser:
This fixes #645 and maybe also #532.


kares:
Thanks Falk, MS code is quite hacky and since there's been issues with merging code without checking for test regressions it will have to wait until I'm able to run tests against SQLServer and verify. You can speed this up by running the suite yourself and confirming it did not cause any additional failures, thanks!

Issue #572, opened 2014-08-06
todo list for 1.4


kares:
## if you're using our master branch (1.4 pre-release) and you do like where it's heading please **[salt the project](https://salt.bountysource.com/teams/arjdbc)**.

AR-JDBC **1.4.x** releases aim to finally provide **performance improvements** that should much up the AR query execution time with MRI (and beyond). Historically optimization has not been done much as most of the time is spent hustling with AR compatibility and supported adapter variability. There's certainly a lot of room for improvement and for most production adapters such as MySQL, PostgreSQL the results should be noticeable. There's also plenty of refactoring (and maybe even some new features) to be done, that has been simply avoided on the 1.3 stable line.
- [ ] performance improvements
  - [x] try to resolve byte ruby String when possible
  - [ ] covert to ruby Time directly from native part
  - [ ] review adapters whether they really need binary/char stream reading
  - [ ] ...
- [x] rewrite JDBCError (support cause compat with 2.1)
- [x] do not wrap runtime exceptions into JDBCErrors
- [x] allow for lazy JDBC connection initialization (should be default for JNDI)
- [x] support disabling `configure_connection` calls using a config value
- [ ] try to log transaction begin/commit happening similar to core
- [ ] review savepoint tracking (avoid map instantiation and/or use internal variable)
  - [ ] SQLite3 now has working savepoint [API](https://github.com/xerial/sqlite-jdbc/blob/master/src/main/java/org/sqlite/jdbc3/JDBC3Connection.java#L265)
- [ ] no longer use (extend) our JDBC/JNDI callbacks ...
- [ ] stop filling in `:adapter_spec` if there's an adapter class
- [ ] stop filling in `:adapter_class` and pass it down as a jdbc_connection arg
- [ ] we shall handle some translate_exception overrides ?!
  e.g. oracle-enhanced does : 

``` ruby
  def translate_exception(exception, message) #:nodoc:
    case @connection.error_code(exception)
    when 1
      RecordNotUnique.new(message, exception)
    when 2291
      InvalidForeignKey.new(message, exception)
    else
      super
    end
  end
```
- [ ] refactor Column(s) to have the same super class as core
- [x] introduce Column impl class convention e.g. `ArJdbc::MySQL::Column`
- [ ] review adapter class names to be JdbcXxx by default ...
- [ ] ~~patch log to print cause's error (maybe try Rails itself)~~
- [x] refactor our AR benchmarking app
[...]

PR #650, opened 2015-05-17
Issue 649: Correct the getConnectionInternal for JNDI Regression

ericgascoine:
Fixes the GetConnectionInternal() function so that any lazy connection source will function correctly.

This seems to have been a simple coding error that was not exposed until it was used with a Connection source that was set to being lazy, and at this point I can only find that being set for JNDI connections.

This corrects it as the flow is allowed to fall through to the existing code below the test that actually forces the creation of a connection if none exists.

### DIFF
--- a/src/java/arjdbc/jdbc/RubyJdbcConnection.java
+++ b/src/java/arjdbc/jdbc/RubyJdbcConnection.java
@@ -3089,7 +3089,7 @@ public class RubyJdbcConnection extends RubyObject {
     private Connection getConnectionInternal(final boolean required) throws SQLException {
         Connection connection = getConnectionImpl();
         if ( connection == null ) {
-            if ( required || ! connected ) {
+            if ( required && ! connected ) {
                 final Ruby runtime = getRuntime();
                 final RubyClass errorClass = getConnectionNotEstablished( runtime );
                 throw new RaiseException(runtime, errorClass, "no connection available", false);


kares:
thanks, but you're mentioning 1.3.16 in #650 yet you fixed master which is slightly refactored (attempting to do lazy JNDI checkouts by default) ... the [1-3-stable](https://github.com/jruby/activerecord-jdbc-adapter/blob/1-3-stable/src/java/arjdbc/jdbc/RubyJdbcConnection.java#L2734-L2741) connection retrieval code is a bit different so it feels like you fixed the "wrong" thing (unless you really tested master out) ...

ericgascoine:
@kares 

Thanks for asking, I actually did test master as well (it's towards the bottom of the notes on the issue 649), and when I validated that Master is the preferred location for the fixes for Rails 42, I fixed that.

I'd actually forgotten to go back and look at the 1.3 stable code, and am now wondering if that's even a good idea.  (But it definitely is broken in a similar fashion).

kares:
@ericgascoine please go back and try to review under 1.3.x ... I am still hesitant to merge (esp. since there's no tests to verify and not sure if I'll be able to reproduce) - guessing there must be a different cause (also wonder how AR 4.2 is involved) since the logic you changed was introduced to support lazy connections and I do not see how the `if` logic is wrong (quite some testing went in to get it right).

p.s. 1.3.x is still meant to receive 4.2 support, it's just that (despite several promises) no one has done it ... yet ... master is primarily performance improvements and as it is it needs [more work and testing](https://github.com/jruby/activerecord-jdbc-adapter/issues/572) with all adapters/drivers supported - thus far from the final release (~ going to be 1-2 pres first).

kares:
sorry, turns out I was looking at something else locally ... put the commit on master https://github.com/jruby/activerecord-jdbc-adapter/commit/bdf9ec24933a790e5ff39677f4d7c0f03266ea46

good catch on this one, thanks 

still interested in resolving this on 1.3 (since you mentioned in #649 it's an issue with a stable release) but it's your call (feel free to close #649) ... a stack-trace would be a start.

ericgascoine:
Thanks @kares 

Yes, I've continued to look at it on 1.3 as per your suggestion, and am getting closer.  Looks like it's related to the use of the connection after the prepare callback in Rails 4.2, which seems not to be  the case in Rails 4.1

Hopefully I'll have another pull request for you.

kares:
Thanks Eric, what  would be very valuable are details of your findings around what's going different around JNDI in 4.2 (reported in an issue if needed).

ericgascoine:
@kares 
Thanks for the note about the full details in #649  When I replied to you previously, I had not got that far.  

I'm still at a bit of a loss as to how to handle this as it's only affecting JNDI connection code (and therefore not a Rails 42 piece of code), but it does indicate a change to the expectations of Rails will affect many people (including more than one of my prior clients)  
In hindsight, I probably should have closed the original issue 649 and opened a new issue rather than changing the title, and can still do so, if it makes it easier for you and the other maintainers with commit privileges.

kares:
yeah, patience usually helps to get through some of the tough AR-JDBC pieces :)

seems like I should be able to reproduce easily due your feedback (e.g. with Trinidad), so I might look into that eventually - AR 4.2 as is not fully supported anyway (1.3.x prints a warning), so I would not worry too much for now. MySQL is the only one that somehow works but I mostly got the tests passing I do not have any clients/app or feedback from production ... and there's probably an underlying refactoring that will need to take place (due AR's new type system) as a base to work with *adapter: jdbc* ... so for now thanks a lot!

PR #653, opened 2015-05-22
 Fix postgres array test for Rails 4.2

prathamesh-sonpatki:
- Defined array? method in ArJdbc::Postgresql::Column module which is
   used in tests for Rails 4.2.
 - Also fixed default value of array in PostgreSQL in tests.


kares:
on 1-3-stable as https://github.com/jruby/activerecord-jdbc-adapter/commit/125f70f86dcb880d4795058b71206b767c8bf327 thanks

Issue #652, opened 2015-05-20
Rails 4.2 under Jruby can't find PGconn class


rafaelfranca:
_From @ninkibah on May 20, 2015 16:39_

I am trying to upgrade our Rails 3.2 app to 4.2.1. So far, I've managed to get to 4.1.6 and I'm almost there. However, I am getting an 
`
unknown constant  ActiveRecord::ConnectionAdapters::PostgreSQL::OID:Bytea::PGconn
`

I suspect it has something to do with the ARJDBC adapter.

As a workaround, I added the following code to a config/initializer:

```ruby
  module ConnectionAdapters
    module PostgreSQL
      module OID # :nodoc:
        class Bytea < Type::Binary # :nodoc:
          def type_cast_from_database(value)
            return if value.nil?
            return value.to_s if value.is_a?(Type::Binary::Data)
            val = super
            return val unless val.starts_with?('\x')
            [val[2..-1]].pack('H*')
          end
        end
      end
    end
```

The last two lines seem to be functionally equivalent to PGconn.unescape_bytea(super).

Question 1: Is this code as good as using the C/Java function which would normally be called? I presume pack is efficiently implemented in both Ruby and JRuby.

Question 2: Is this code worth doing a pull request for, or is it better to have the dependency on PGconn?

_Copied from original issue: rails/rails#20229_

rafaelfranca:
> or is it better to have the dependency on PGconn

We already depend o PGconn, so we should just use it normally here. Maybe we should change `PGconn.unescape_bytea(super)` to `::PGconn.unescape_bytea(super)`?

rafaelfranca:
_From @prathamesh-sonpatki on May 20, 2015 17:44_

PG gem is not used with JRuby, This code needs to be overridden in ActiveRecord JDBC side.

rafaelfranca:
I see. I think we see where we use `PGconn` inside Rails adapter directly and isolate that in internal adapter calls so the JRuby adapter only override these isolated methods. @ninkibah do you want to work on this?

rafaelfranca:
hmm, but I think the type objects should not know about the connection. So maybe your patch is fine. @sgrif WDYT?

rafaelfranca:
_From @sgrif on May 20, 2015 17:55_

[...]

PR #655, opened 2015-06-07
Added minimal implementation of PGconn with unescape_bytea

### LINKED ISSUE #651: Problem reading postgresql table with binary/bytea column
I'm trying to upgrade our project to Rails 4.2.1, and when I try to load data from a table with a bytea column, created using t.binary in the migration file, I get an error:

`uninitialized constant ActiveRecord::ConnectionAdapters::PostgreSQL::OID::Bytea::PGconn`

I have tried this with both master and 1.3.6, with the same problem. A sample stack trace is:

```
        from /home/joconnor/.rvm/gems/jruby-1.7.20@xpesa/gems/activerecord-4.2.1/lib/active_record/connection_adapters/postgresql/oid/bytea.rb:9:in `type_cast_from_database'
        from /home/joconnor/.rvm/gems/jruby-1.7.20@xpesa/gems/activerecord-4.2.1/lib/active_record/attribute.rb:104:in `type_cast'
        from /home/joconnor/.rvm/gems/jruby-1.7.20@xpesa/gems/activerecord-4.2.1/lib/active_record/attribute.rb:42:in `original_value'
        from /home/joconnor/.rvm/gems/jruby-1.7.20@xpesa/gems/activerecord-4.2.1/lib/active_record/attribute.rb:37:in `value'
        from /home/joconnor/.rvm/gems/jruby-1.7.20@xpesa/gems/activerecord-4.2.1/lib/active_record/attribute_set.rb:31:in `fetch_value'
        from /home/joconnor/.rvm/gems/jruby-1.7.20@xpesa/gems/activerecord-4.2.1/lib/active_record/attribute_methods/read.rb:93:in `_read_attribute'
        from /home/joconnor/.rvm/gems/jruby-1.7.20@xpesa/gems/activerecord-4.2.1/lib/active_record/attribute_methods/read.rb:87:in `read_attribute'
        from /home/joconnor/.rvm/gems/jruby-1.7.20@xpesa/gems/activerecord-4.2.1/lib/active_record/attribute_methods.rb:305:in `attribute_for_inspect'
        from /home/joconnor/.rvm/gems/jruby-1.7.20@xpesa/gems/activerecord-4.2.1/lib/active_record/core.rb:444:in `inspect'
        from /home/joconnor/.rvm/gems/jruby-1.7.20@xpesa/gems/activerecord-4.2.1/lib/active_record/core.rb:442:in `collect'
        from /home/joconnor/.rvm/gems/jruby-1.7.20@xpesa/gems/activerecord-4.2.1/lib/active_record/core.rb:442:in `inspect'
        from /home/joconnor/.rvm/gems/jruby-1.7.20@xpesa/gems/railties-4.2.1/lib/rails/commands/console.rb:110:in `start'
 ```

This was not a problem with Rails 4.1.6 or earlier.

Has anyone else seen this problem. Google was not helpful in this regard.

The fix is to implement PGconn.unescape_bytea, which is what I may end up doing.

donv:
To fix #651 and #652 .

### DIFF
--- a/lib/arjdbc/postgresql/base/pgconn.rb
+++ b/lib/arjdbc/postgresql/base/pgconn.rb
@@ -1,0 +1,5 @@
+class PGconn
+  def self.unescape_bytea(escaped)
+    String.from_java_bytes Java::OrgPostgresqlUtil::PGbytea.toBytes escaped.to_java_bytes
+  end
+end
--- a/lib/arjdbc/postgresql/oid_types.rb
+++ b/lib/arjdbc/postgresql/oid_types.rb
@@ -4,6 +4,7 @@ module ArJdbc
   module PostgreSQL
 
     if AR42_COMPAT
+      require 'arjdbc/postgresql/base/pgconn'
       require 'active_record/connection_adapters/postgresql/oid'
     else
       require 'arjdbc/postgresql/base/oid'


donv:
Before:

    380 tests, 608 assertions, 4 failures, 21 errors, 2 pendings, 72 omissions, 0 notifications

After:

    380 tests, 639 assertions, 4 failures, 17 errors, 2 pendings, 72 omissions, 0 notifications

4 errors fixed.  OK to merge?

prathamesh-sonpatki:
:+1:

kares:
Thanks Uwe, looks good however with jruby-pg around it's quite ugly to pollute their name-space ... we did moneky-patch AR previously so would you mind if I change that ?

also I'm not sure why you put up the 1.4 target ... I'd like to put it on 1-3-stable and then merge here so that it lands in next stable release (for 1.4 I would also move this code into native), thoughts?

donv:
Hi @kares !

Thanks for looking at this.  Feel free to change as you see fit :smile: and put on the 1-3-stable as well.  I was unsure if 1-3-stable targeted AR 4.2.

I got an error using schema_plus without any implemented PGconn because the OID implementation for PostgreSQL is loaded then.  Not sure if this is intended, but it seems ARJDBC aims to use as much standard AR as possible, and AR uses PGconn directly.

kares:
oh, you guys merged that already - well just go for it than - I do not have time to "play" around AR 4.2 compatibility hacks just be aware when introducing another :) ... planning a 1.4 pre-release from master.

kares:
@donv OID was completely re-invented previously, AR "promised" mixins will be shareable ...but it's not :(

donv:
I'd like everyone to play nice together, but it seems jruby-pg and ARJDBC are exclusive, right?  They should not be used together?  AR+ARJDBC or AR+jruby-pg, right?

kares:
why would they be exclusive? ... I mean for AR they might be, but pg is not made solely for AR ...

kares:
... you can hide the `PGConn` under `OID` module (for now) ... it should work since there's no `::PGConn`

donv:
Oh.  I'm learning, then :)  How would you use jruby-pg with ARJDBC?

I'll move the PGconn class below the ActiveRecord::ConnectionAdapters::PostgreSQL::OID module.

kares:
> I mean for AR they might be, but pg is not made solely for AR ...

... one can access DBs in different ways (along side with AR) - which is probably uncommon but possible

PR #663, opened 2015-07-07
track upstream: remove TRUE_VALUES, use !FALSE_VALUES.include? instead

djellemah:
track this activerecord-5 commit

https://github.com/rails/rails/commit/a502703c3d2151d4d3b421b29fefdac5ad05df61


kares:
thanks, while this looks good to me ... we'll likely need to support older AR releases at the same time.
as we might need to mirror the pre-5 behaviour ... not 100% sure but users might expect us to do so.

kares:
picked this on master with https://github.com/jruby/activerecord-jdbc-adapter/commit/783bd8abe38ef31752bdac5ff3418e2592ae9a78 ... to provide backwards compatibility, thanks!

PR #671, opened 2015-07-18
Allow reading "infinity" and "BC timestamps

donv:
Reverted tests mistakenly adjusted to AR42

Tests are green for PostgreSQL locally for AR41 and AR42.


donv:
Taking a look at the remaining PostgreSQL+AR42 errors now

PR #673, opened 2015-07-20
Use JDBC instead of PostgreSQL prepared statement bind marker

donv:


donv:
Hi @kares !

Thanks for looking at this :smile: 

This overrides the AREL gem like we do for Derby.  Arel has a special patch for PostgreSQL to convert the bind markers:

https://github.com/rails/arel/blob/master/lib/arel/visitors/postgresql.rb#L39

This messes up the JDBC prepared statements since JDBC uses "?" for bind markers.

Is this not the way to do it?

donv:
I used a separate file name to not collide with AREL source.

donv:
Reduces test failures from 155 to 35 for PostgreSQL with prepared statements.

kares:
yes but loading the "arel" version will collide since its the same name ... 2 possibilities : 

* `class JdbcPostgreSQL < Arel::Visitors::PostgreSQL`
* or patch the AREL bind marker line/method to do nothing

1. is better but 2. might be needed - as things might blow-up easily since AREL/Rails used to assume class-names in a map thus `['PostgreSQL']` might no longer work (not sure where it blows up) ...

kares:
excellent that you found the cause for those "sudden" PS failures ... just keep in mind AREL is considered internal to Rails and thus it changes a lot (between versions) - it's quite a thin ice to walk on.

donv:
@kares do you think I should undefine  Arel::Visitors::PostgreSQL#visit_Arel_Nodes_BindParam or make it call ```super```?

kares:
@donv +1 for undef or re-def once again to do "correct" thing with a NOTE: why :) ... if it's working across versions. but it's completely your call I do not have the sufficient context as I'm just looking at your diffs ...

donv:
OK, changed to use ```remove_method``` to enable the super class logic.  That is better.  I appreciate all your comments @kares :smile: 

Awaiting the travis build.  OK to merge?  Give me a :+1:  if so :smiley:

kares:
:green_heart:

PR #674, opened 2015-07-21
Store decimals without scale without loss in PostgreSQL

donv:
Mark the hex bit test as failing with prepared statements for now


kares:
have you considered doing the .jar commits separate? its easier to merge that way, also if you're planning more work on the Java side just do a single .jar commit for multiple .java changes (I usually only compile the .jar once/twice between 1.3 releases - I do test .java changes with adapter_java.jar commited on my dirty fork).

.git repo is over 100MB which is fine but we'll constantly adding 1M+ jars for the drivers already, although the adapter_java.jar is ~ 150k ... 10 small .java changes seem not unusual the way you roll :)

its very non ideal and "dirty" (that is why master is hacked to build auto-magically) but we'll paying the price of a huge download with `gem 'ar-jdbc' :git => ...` ;(

donv:
Hi @kares !

I'll do the jar commits separate from now on, but in the same push, so they only generate one travis build.

I need to push the jar to get a useful build in travis.  Not sure how to work around that.

    gem 'activerecord-jdbc-adapter', git: ...

...should do a "shallow" checkout without history, so should be quick enough.  All who use the git repo should be prepared for the download.

BTW:  OK to merge?  I know I don't need approval, but it is so useful to get a second opinion on things.

After this merge, PG + PS has only **ONE** failure.

kares:
ok than ... did not know bundler does a shallow clone (actually did not know that there's such thing, thx!). 

:green_heart: :watermelon: :green_heart:

PR #676, opened 2015-07-22
Retrieve binary data as an ActiveRecord::Type::Binary::Data when using prepared statements

donv:
This avoid extra unescape and is symmetrical (Data in, Data out).

Huge impact on source, so I don't like this solution.  **Do not merge!**

Anyone see a better way?


kares:
quite HUGE and will likely break existing AR-JDBC extensions - NOT merge material for **1-3-stable** if an API change is needed it should instead happen in **1.4** (should look at options of patching AR::Type)

PR #678, opened 2015-07-23
Store binary data as-is for PostgreSQL

### LINKED ISSUE #675: Binaries are unescaped twice with AR42 + PostgreSQL + prepared statements
Failing test:  https://travis-ci.org/jruby/activerecord-jdbc-adapter/jobs/71944362#L624

Initial investigation shows that the binary content comes **unescaped** from RubyJdbcConnection (@connection.execute_query(sql, binds)) to ActiveRecord::ConnectionAdapters::JdbcAdapter#exec_query when using prepared statements:

https://github.com/jruby/activerecord-jdbc-adapter/blob/1-3-stable/lib/arjdbc/jdbc/adapter.rb#L449

When not using prepared statements, the result **IS** escaped.

We need to investigate how to avoid unescaping the result when using prepared statements.

I have not found an obvious place to do that, yet, and all help is appreciated.

donv:
Use the "E" prefix to mark the input as escaped binary data, and do not unescape when reading the data back from the DB.

Fixes #675

### DIFF
--- a/lib/arjdbc/postgresql/adapter.rb
+++ b/lib/arjdbc/postgresql/adapter.rb
@@ -918,7 +918,7 @@ module ArJdbc
     def _quote(value)
       case value
       when Type::Binary::Data
-        "'#{escape_bytea(value.to_s)}'"
+        "E'#{escape_bytea(value.to_s)}'"
       when OID::Xml::Data
         "xml '#{quote_string(value.to_s)}'"
       when OID::Bit::Data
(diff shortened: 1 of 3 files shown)


kares:
heh, najs trick from your pocket! ... 'E' is a Postgre's thing, I guess? :green_heart:  :cupid: :green_heart:

donv:
Yes, 'E' is connected to 'bytea' for PostgreSQL.  Old fashioned escaping with octal escape sequences did not require any prefix, but HEX encoded escape sequence requires the 'E' prefix.  Without the 'E' prefix, the string is stored directly without being unescaped as a hex sequence.

This was introduced in PG 8.2.  How old versions of PG do we support?  The README does not say anything about that.

kares:
we sure supported older than AR itself ... there's semi-support for Amazon's Redhift, but I mostly "blind" fixed some reported requests (without running against Redshift), which is based on PostgreSQL 8.0.2 ...

PR #679, opened 2015-07-24
Use standard AREL for AR42 with H2/HSQLDB

donv:
H2/HSQLDB support standard LIMIT and OFFSET now.


kares:
do you happen to know if they always did (if not since what versions they do) ... as I recall they did not?

donv:
This doc http://www.hsqldb.org/doc/guide/ch09.html is dated 2007/02/19, so it is pretty old.

PR #680, opened 2015-07-24
Fix LIMIT and OFFSET for H2 and HSQLDB with AR 4.2

donv:
This actually fixes this feature for H2 and HSQLDB  for older versions of AR as well.  When only OFFSET was set, the result was wrong.


donv:
Argh!  PR for master instead of 1-3-stable...closing.

PR #681, opened 2015-07-24
Fix H2/HSQLDB LIMIT and OFFSET with AR 4.2

donv:
Adds a test for LIMIT and OFFSET including having an OFFSET without a LIMIT.  This was broken for older versions of AR as well, but fixed now.


kares:
:green_heart: :cool: :purple_heart:

PR #683, opened 2015-07-26
Fix travis-ci builds with Ruby --1.8 mode

donv:
Activated --1.8 mode and made everything green :smile: 

@kares ok to merge?


donv:
Everything is green, so I merge :smile:

PR #693, opened 2015-11-19
Minor whitespace and empty comment cleanup

deivid-rodriguez:
I removed this while browsing the file.


kares:
OK, picked on 1-3-stable

PR #694, opened 2015-11-25
Fixed MySQL primary key issue with MySQL >= 5.7.3

Keeguon:
Self explanatory (also in sync w/ current ActiveRecord code).

### DIFF
--- a/lib/arjdbc/mysql/adapter.rb
+++ b/lib/arjdbc/mysql/adapter.rb
@@ -92,7 +92,7 @@ module ArJdbc
     def self.emulate_booleans=(emulate); @@emulate_booleans = emulate; end
 
     NATIVE_DATABASE_TYPES = {
-      :primary_key => "int(11) DEFAULT NULL auto_increment PRIMARY KEY",
+      :primary_key => "int(11) auto_increment PRIMARY KEY",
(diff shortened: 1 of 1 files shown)


kares:
Thanks @Keeguon, is the change backwards compatible since AR-JDBC runs against older AR versions?

Keeguon:
@kares Since it's mainly how the primary key is handled it has no impact moreover IMO it wasn't making a whole lot of sense for a primary key to be defaulted to NULL in MySQL. The change was made in AR 4.1 which means older versions don't support MySQL >= 5.7.3 natively and a monkey patch must be applied. Also, I did tests yesterday and also backported this to the 1.3 branch on my fork (if you want me to do a PR for this branch as well tell me) because I needed that for a project.

kares:
Thanks for confirming. If you want to see it in next 1.3 release its preferable to open this against 1.3-stable branch - no need for this in that case as it will merge to master later. If you do so, please link the other PR here - your comment might be valuable for others.

Keeguon:
Here's the PR for the 1-3-stable branch: https://github.com/jruby/activerecord-jdbc-adapter/pull/695

kares:
will merge over from 1-3-stable (due #695), thanks for the contribution.

PR #698, opened 2015-12-05
Remove comment about pg_params going away

jkutner:
Pretty please <3 <3 <3


kares:
heh, well its kind of jdbc adapter specific and one can set the `:url` is it convenient to have around?
thought at some point older AR supported this going down to PG ... [currently no pg_params accepted](https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L23-L44)

jkutner:
It's important for a few PG JDBC URL parameters like `sslmode=require` and `ssl=true` and `sslfactory=org.postgresql.ssl.NonValidatingFactory`. When using `DATABASE_URL`, as apps should, it's difficult to add these in any other way. Furthermore, by keeping it specific to PG, rather than just having something generic like `jdbc_params`, it allows you to easily switch DB vendors.

kares:
OK, than - keeping it up for generation H!

PR #699, opened 2015-12-11
support for DB2 big integer (BIGINT) data type

boskowski:
I made this change in order to be able to create fields of type BIGINT. Without, the adapter adds a limit (e.g. ```BIGINT(19)```), which is inappropriate and causes failure.


kares:
https://github.com/jruby/activerecord-jdbc-adapter/commit/2d591efb892dfdbec856bd634ab138c2f2297f7a on 1-3-stable, thanks

PR #714, opened 2016-03-02
Enable DDL transactions for Derby

donv:
DDL (migration) transactions used to be enabled for Derby.  Not sure when/how/why it was disabled.

OK, to merge?


donv:
@kares : Should I commit to 1.3 branch instead?

kares:
@donv yep, go ahead and commit into **1-3-stable** instead, will get merged over

donv:
Closing and committing to 1-3-stable instead

PR #719, opened 2016-04-06
Initial support for ActiveRecord 5.0 beta

mattclar:
This can run some migrations and do some basic CRUD stuff, it has not been thoroughly tested and if you did you would likely find that it doesn't work great, but it is a start.

It should go someway to solving issue #700


kares:
thanks, as originally noted when being presented with this code - its hard to follow up from where its left.
we either need more than "some" or something less of an "all in-one" commit (so that someone could pick it up from there) - try running our tests and fix things if you're interested in helping. closing as there are conflicts by now and people assume as if **ar-50** is our official branch 5.0 support branch :crying_cat_face:

PR #723, opened 2016-04-28
fixed insert into partitioned table

DzmitryNikitsin:


kares:
Thanks Dzmitry! could we have a failing test-case here or at least a mention how to reproduce?

DzmitryNikitsin:
yes, just create partitioned table, let's say customers will be partitioned by country_id. 

Customer.create(name: 'test', country_id: 1)

you will get error, because customers partitioned table in postgres, is it good ?

kares:
went without a test and picked your commit into 1.3.22, thanks!

Issue #731, opened 2016-06-17
When migrating db: NotImplementedError: #views is not implemented 


mattclar:
I am using the rails-5 branch which is working ok with some workarounds, however I went to run a migration today to add a column to a table, however I ran into a not implemented error!
Not sure if anyone has a workaround for this at all

```
 == 20160617012150 Addemailtopatients: migrating =============================== 
 -- add_column(:patients, :email, :string) 
    -> 0.0267s 
    -> 0 rows 
 == 20160617012150 Addemailtopatients: migrated (0.0331s) ====================== 
 rails aborted! 
 NotImplementedError: #views is not implemented 
 /usr/local/bundle/gems/activerecord-5.0.0.rc1/lib/active_record/connection_adapters/abstract/schema_statements.rb:60:in `views' 
 /usr/local/bundle/gems/activerecord-5.0.0.rc1/lib/active_record/schema_dumper.rb:92:in `tables' 
 /usr/local/bundle/gems/activerecord-5.0.0.rc1/lib/active_record/schema_dumper.rb:37:in `dump' 
 /usr/local/bundle/gems/activerecord-5.0.0.rc1/lib/active_record/schema_dumper.rb:21:in `dump' 
 /usr/local/bundle/gems/activerecord-5.0.0.rc1/lib/active_record/railties/databases.rake:253:in `block in (root)' 
 /usr/local/bundle/gems/activerecord-5.0.0.rc1/lib/active_record/railties/databases.rake:252:in `block in (root)' 
 /usr/local/bundle/gems/bugsnag-4.1.0/lib/bugsnag/rake.rb:12:in `execute_with_bugsnag' 
 /usr/local/bundle/gems/activerecord-5.0.0.rc1/lib/active_record/railties/databases.rake:66:in `block in (root)' 
 /usr/local/bundle/gems/bugsnag-4.1.0/lib/bugsnag/rake.rb:12:in `execute_with_bugsnag' 
 /usr/local/bundle/gems/activerecord-5.0.0.rc1/lib/active_record/railties/databases.rake:59:in `block in (root)' 
 /usr/local/bundle/gems/bugsnag-4.1.0/lib/bugsnag/rake.rb:12:in `execute_with_bugsnag' 
 /usr/local/bundle/gems/railties-5.0.0.rc1/lib/rails/commands/rake_proxy.rb:13:in `block in run_rake_task' 
[...]

PR #732, opened 2016-06-17
Add views to postgres adapter

mattclar:


mattclar:
I believe this fixes the issue #731

PR #733, opened 2016-06-17
Adds citext to Postgres AR42 Native database types.

zachabrahams:
In the current version of the postgres adapter,  Rails has problems generating a schema.rb for a table that uses a citext. Instead of the table's schema, it generate:

```
#  Could not dump table "whatever_table_name_is" because of following StandardError
#   Unknown type 'citext' for column 'email'
```

The cause is that valid_type? is coming back false for citext since it isn't included in NATIVE_DATABASE_TYPES.  

Let me know if you'd like me to add a test (and where a good place to add one would be!)


kares:
picked into AR-JDBC 1.3.22, thanks for the contribution!

PR #740, opened 2016-08-17
Allow RubyArray and all subclasses to pass this check.

headius:
See jruby/jruby#4080

### DIFF
--- a/src/java/arjdbc/jdbc/RubyJdbcConnection.java
+++ b/src/java/arjdbc/jdbc/RubyJdbcConnection.java
@@ -2146,7 +2146,7 @@ public class RubyJdbcConnection extends RubyObject {
         for ( int i = 0; i < binds.size(); i++ ) {
             // [ [ column1, param1 ], [ column2, param2 ], ... ]
             Object param = binds.get(i); IRubyObject column = null;
-            if ( param.getClass() == RubyArray.class ) {
+            if ( param instanceof RubyArray ) {
                 final RubyArray _param = (RubyArray) param;
                 column = _param.eltInternal(0); param = _param.eltInternal(1);
             }


elthariel:
That'd be great to merge this to the rails-5 branch as well :)

enebo:
@elthariel cherry-picked at 5468b6f.  Thanks for pointing that out. 

@kares should we consider a merge do you think of other commits to rails-5?

kares:
@enebo your call, I somehow lost track what rails-5 is based off. anyways you should be good if you're off 1-3-stable - master's "refactorings" are unfinished and not released (no plans really to work on it atm).

elthariel:
Thanks for the impressive reactivity !

enebo:
@kares well call me dumb but I never evaluated the difference between 1-3-stable and master.  master is basically a good portion of the way to satisfying the last bit of Rails 5 work (namely that execute handles all queries/updates/inserts/...).  So I may look this over and at least steal some stuff.  sqlite3 adapter is 5E/5F on rails-5 and really 2 of those are missing loading Arel in tests.  So I am slowly getting this green.

kares:
NP - well give it a try if you can but there's certainly some regressions (on master) I haven't looked into :(
... glad someone is finally spending some quality time on AR-JDBC :1234: :five: :five: :five: !

enebo:
@kares my other issue atm in using what we have is lots of deprecations on using things like quote so although what we have might pass tests it won't work for long so some of your work I think will be more of a starting point.

kares:
@enebo honestly I did not remove that much - not sure its ben a while (mostly some really old AR 2.x relics) have concentrated on cleaning up the internal APIs ... for better speed and some less tech.debt.

PR #743, opened 2016-08-25
Repair postgres array support

mattclar:
They way activerecord implemented array support changed from ruby based to a C based implementation to apparently speed it up, this reverts that change for our purposes for obvious reasons of not supporting C
Also note that I discovered a gem called pg_array_parser that implements a C based and a Java based speedier method.

Another tick along the line of resolving #708

### DIFF
--- a/lib/arjdbc/postgresql/base/oid.rb
+++ b/lib/arjdbc/postgresql/base/oid.rb
@@ -403,6 +403,9 @@ module ActiveRecord
         register_type 'json', OID::Json.new
         register_type 'jsonb', OID::Jsonb.new
         register_type 'ltree', OID::Identity.new
+        
+        register_type 'array', OID::Array.new
+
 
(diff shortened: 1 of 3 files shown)


kares:
thanks, but this isn't a good idea for maintenance (also very verbose yet little telling commits). 
doubt pg gem has the whole AR::Type implemented, thus AR-JDBC should match this closely and provide a similar (native) method as AR uses from PG's API.

mattclar:
Sorry for the verbose commits, I didn't realise they would flow through to the pull request (noob)!
Also I agree this is a change that should be made on the activerecord project as they broke this in the first place and it is more easily maintained there!
I have opened an issue to this affect https://github.com/rails/rails/issues/26338

kares:
Thanks Matt, its unfortunate they aren't at least willing to provide an indirection from the PG API :disappointed:

PR #746, opened 2016-09-05
Update to postgres' jdbc driver 9.4-1209

### LINKED ISSUE #710: not compatible with postgres-jdbc driver 9.4.1207
due changing (internal?) API of the driver that we're using ... no `Jdbc4Array` class

https://github.com/pgjdbc/pgjdbc/compare/REL9_4_1206...REL9.4.1207

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/30564581-not-compatible-with-postgres-jdbc-driver-9-4-1207?utm_campaign=plugin&utm_content=tracker%2F136963&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F136963&utm_medium=issues&utm_source=github).

MSNexploder:
Bumped postgres jdbc driver to latest version.

Removed usage of internal (and no longer existing) Jdbc4Array class.
This should fix #710.


kares:
thanks - this is probably good for (unreleased) master but for stable it would need to handle backwards compatibility since it's common for people to be using the older version. there needs to be a clear (if possible painless) transition.

MSNexploder:
Good point!  I'm looking into it - shouldn't take too much hackery to support older versions.

MSNexploder:
Sorry for not coming back to this sooner.
I'm pretty sure with some hackery supporting older versions should be possible.

Given that there is already a new major driver version (which drops support for PostgreSQL 8.2 and below) and the limited resources i'm unsure which driver versions should be support in which adapter version.

MSNexploder:
Just took another shoot at this and while I'm fairly certain that I was able to make the change backwards compatible, I realised there is no test which actually executes #setStringParameter in a way that triggers the specific code branch and I'm unable to create a test triggering this.
Which means I'm unsure if the code is doing the right thing or is needed at all.

My current take would be to ditch this pull request and keep the 1.3 branch on the old "9.4.1206" driver avoiding any major breaks and creating a new pull request against master including the (then breaking) update to the latest "42.1.x" PostgreSQL driver.

@kares any thoughts?

src/java/arjdbc/postgresql/PGDriverImplementation.java
```diff
@@ -164,6 +166,28 @@ private static ByteList formatInterval(final Object object) {
...
     }
 
+    private static Class<?> pgArrayClass;
+    private static Class<?> pgArrayClass(final ThreadContext context) {
+        final Class<?> clazz = pgArrayClass;
+        if (clazz == null) {
+            final Ruby runtime = context.getRuntime();
+            try {
+                // try loading official PgArray Class (since postgresql 9.4)
+                pgArrayClass = runtime.getJavaSupport().loadJavaClass("org.postgresql.jdbc.PgArray");
```
kares: it might not come from JRuby's class-loader

src/java/arjdbc/postgresql/PGDriverImplementation.java
```diff
@@ -164,6 +166,28 @@ private static ByteList formatInterval(final Object object) {
...
+                return pgArrayClass;
+            }
+            catch (ClassNotFoundException ex) {
+                try {
+                    // load older (internal) Jdbc4Array Class (9.2 and earlier)
+                    pgArrayClass = runtime.getJavaSupport().loadJavaClass("org.postgresql.jdbc4.Jdbc4Array");
```
kares: same here ... although I am not sure if it matters - it probably will find the class assuming parent delegation is on

src/java/arjdbc/postgresql/PGDriverImplementation.java
```diff
@@ -190,8 +214,17 @@ public boolean setStringParameter(final ThreadContext context,
             if ( sqlType != null ) {
                 if ( PostgreSQLRubyJdbcConnection.rawArrayType == Boolean.TRUE && arrayLike(sqlType) ) {
                     final int oid = PostgreSQLRubyJdbcConnection.oid(context, column);
-                    Jdbc4Array valueArr = new Jdbc4Array(connection.unwrap(BaseConnection.class), oid, valueStr);
-                    statement.setArray(index, valueArr); return true;
+                    final Class<?> clazz = pgArrayClass(context);
+                    try {
+                        Array valueArr = (Array) clazz
```
kares: this is good ... but the driver we're compiling against should be implemented directly with a reflection fallback if using and older driver. detection should happen once (e.g. in a static final flag).

src/java/arjdbc/postgresql/PGDriverImplementation.java
```diff
@@ -164,6 +170,32 @@ private static ByteList formatInterval(final Object object) {
...
+            }
+            catch (ClassNotFoundException e) { /* should never be reached */ }
+        }
+        return clazz;
+    }
+
+    private static boolean shouldUseInternalPGArrayClass() {
+        try {
+            // try loading official PgArray Class (since postgresql 9.4)
+            Class.forName("import org.postgresql.jdbc.PgArray");
```
kares: this obviously wasn't tested for backwards compatiblity, much :disappointed:

MSNexploder: This is heavily WiP and was indeed not tested at all - i merely committed because i had to leave.
This was never intended to get actually pushed to GitHub...

Sorry for wasting your time 😔

perlun:
@MSNexploder Will you be able to finish this?

MSNexploder:
@perlun eventually yes - i guess.

Sadly right now i'm swamped with work.
Feel free to chime in if you like, i would really appreciate it.

perlun:
> Feel free to chime in if you like, i would really appreciate it.

Not at the moment, just curious. I know the feeling of not having enough time. 😞

MSNexploder:
Finally got time to work on this.

Still not sure about how to (properly) test  the changes to the super specific edge case handling in PGDriverImplementation#setStringParameter.

kares:
mostly done and explained at https://github.com/jruby/activerecord-jdbc-adapter/issues/710#issuecomment-331736857 ... feel free to test out from master, 
jdbc-postgres gem releases to follow later

MSNexploder:
Just did a quick test from master. Looks good to me.
Sorry for wasting your time and ultimately failing you on this one 😔

Issue #755, opened 2016-10-12
Java::JavaLang::NullPointerException on RubyJdbcConnection.extractColumns


victormartins:
Hello everyone, 
I've been fighting an integration problem with a SqlAnywhere database and currently I'm stuck with a `Java::JavaLang::NullPointerException` in activerecord-jdbc-adapter.

I've started from an application running the following configuration, that has no problems using the DB.
`jRuby 1.7.15`, 
`Rails 3.2.15` 
`activerecord-jdbc-adapter 1.2.91`
`activerecord-sqlanywhere-jdbc-in4systems-adapter 1.0.13`

However, updating to this configuration I now have the exception:
`jRuby 9.1.5`
`rails 3.2.22.5`
`activerecord-jdbc-adapter 1.3.21`
`activerecord-sqlanywhere-jdbc-in4systems-adapter 1.0.13`

When trying to execute a INSERT query, I can see that the query is executed correctly, that data gets set in the DB, but I immediately get the exception bellow.
I've been trying to find a way to compile a new jar file with some log to debug this, but I've been failing so far. I'm trying to understand what is causing this error, but I can't pinpoint the change that broke it.
Can someone please try to highlight what can cause such an exception? 

Thank you.

```
     Java::JavaLang::NullPointerException:
     # arjdbc.jdbc.RubyJdbcConnection.extractColumns(RubyJdbcConnection.java:3444)
     # arjdbc.jdbc.RubyJdbcConnection.mapQueryResult(RubyJdbcConnection.java:989)
     # arjdbc.jdbc.RubyJdbcConnection.access$200(RubyJdbcConnection.java:96)
     # arjdbc.jdbc.RubyJdbcConnection$9.call(RubyJdbcConnection.java:954)
     # arjdbc.jdbc.RubyJdbcConnection$9.call(RubyJdbcConnection.java:947)
     # arjdbc.jdbc.RubyJdbcConnection.withConnection(RubyJdbcConnection.java:3486)
     # arjdbc.jdbc.RubyJdbcConnection.withConnection(RubyJdbcConnection.java:3467)
     # arjdbc.jdbc.RubyJdbcConnection.executeQuery(RubyJdbcConnection.java:947)
     # arjdbc.jdbc.RubyJdbcConnection.execute_query(RubyJdbcConnection.java:888)
[...]

PR #757, opened 2016-10-26
Added long varchar to type converter for legacy sybase DB

victormartins:
We are using a legacy Sybase DB and we have the long varchar type in use.
The "text" data type presented on line 35 is an alias for that type in Sybase, but for legacy codebases
the long varchar will probably be in use as it happens for us.

### DIFF
--- a/lib/arjdbc/jdbc/type_converter.rb
+++ b/lib/arjdbc/jdbc/type_converter.rb
@@ -34,6 +34,7 @@ module ActiveRecord
         :text        => [ lambda {|r| TEXT_TYPES.include?(r['data_type'].to_i)},
                           lambda {|r| r['type_name'] =~ /^text$/i},     # For Informix
                           lambda {|r| r['type_name'] =~ /sub_type 1$/i}, # For FireBird
+                          lambda {|r| r['type_name'] =~ /^long varchar$/i}, # Legacy Sybase/SqlAnywhere
                           lambda {|r| r['type_name'] =~ /^(text|clob)$/i},
                           lambda {|r| r['type_name'] =~ /^character large object$/i},
                           lambda {|r| r['sql_data_type'] == 2005}],


victormartins:
Hello @enebo , do you think we can get this in, maybe in a 1.3.22 version?
Cheers.

enebo:
@victormartins I think @kares should look at this one and hopefully release afterwards :)

I don't know (or remember) enough on mapping to know how likely that would be to affect another db but I think casually it is longer than other expressions so another db would only run afoul if it also happened to have that string (and then it probably would be this type too).

kares:
might effect others using `adapter: jdbc`, although it should be fine but if it breaks it might get revert-ed.
also isn't the Sybase 'legacy' kind of broken, it still needs another hack, right?

victormartins:
@kares In my view, if nobody else worried about `long varchar` so far or they are hacking it locally as we were doing or they don't use it.
For Sybase it is a legit type, so it would be good to have it I think.

kares:
@victormartins was asking about that other Sybase part. if we're including support for a "neglected" DB but it still doesn't work much out-of-the-box its a bit sloppy. any chance to get reconsider the other part as well?

victormartins:
I'm sorry @kares I don't think I understood your question.
What is the other part you mention?
We need no other changes on this Gem, the code that comes from here works fine, it was just this data type that was missing.
Thank you for merging it, now we can remove the monkey patch.

kares:
you mentioned using a snippet in https://github.com/jruby/activerecord-jdbc-adapter/issues/755 ... but no brainer on my end, esp. as you say to not need other changes. the merged piece is going to be part of ARJDBC **1.3.22**.

victormartins:
Hey @kares thank for making this available on 1.3.22.
Now I see what you are mentioning.
After further development and experience with the jdbc adapter I believe that the idea I mention on https://github.com/jruby/activerecord-jdbc-adapter/issues/755 is wrong.
Because we are using an old patched ugly version of the sqlanywhere adapter, I got confused by a monkey patch to the **exec_query** API of jdbc-adapter. We were calling that method and sending other SQL statements besides SELECT, like INSERT. What I did was a little class that will check the SQL statement being sent, and then choose the right jdbc-adapter API for it. But this is solving a problem in our code base, it's not a problem with the jdbc-adapter I believe. 

Legacy... Legacy everywhere!

PR #760, opened 2016-11-08
Support SSL options via environment vars

kendagriff:
Perhaps this isn't the most well developed code, but it became a giant sticking point for me while on Heroku.

I worked with Heroku Support for a period of three hours, and neither of us could figure out how to get `activerecord-jdbc-adapter` to support SSL in a Rails 3.2 app (note Rails 3.2 does not support ENV(...) in database.yml).

My only recourse was this patch that made the library aware of environment vars.

Did I miss something?

### DIFF
--- a/lib/arjdbc/postgresql/connection_methods.rb
+++ b/lib/arjdbc/postgresql/connection_methods.rb
@@ -31,7 +31,7 @@ ArJdbc::ConnectionMethods.module_eval do
     end
     sslmode = config.key?(:sslmode) ? config[:sslmode] : config[:requiressl]
     # NOTE: makes not much sense since this needs some JVM options :
-    # sslmode = ENV['PGSSLMODE'] || ENV['PGREQUIRESSL'] if sslmode.nil?
+    sslmode = ENV['PGSSLMODE'] || ENV['PGREQUIRESSL'] if sslmode.nil?
     unless sslmode.nil? # PG :sslmode - disable|allow|prefer|require
       # JRuby/JVM needs to be started with :
       #  -Djavax.net.ssl.trustStore=mystore -Djavax.net.ssl.trustStorePassword=...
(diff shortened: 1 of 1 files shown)


jkutner:
These [env vars](https://www.postgresql.org/docs/current/static/libpq-envars.html) aren't honored by the [Postgres JDBC adapter](https://github.com/pgjdbc/pgjdbc), which is really a shame. So this patch actually brings the AR JDBC gem in-line with the native PG client, which does support the env vars.

kares:
these might have been in before but they still required some JVM option setting, that is likely the reason I commented out the first part ... `sslmode = ENV['PGSSLMODE'] || ENV['PGREQUIRESSL'] if sslmode.nil?`
is it working now just by setting these with the official driver?

kares:
also, AR-JDBC is (still) not compatible with latest 9.4 driver ... https://github.com/jruby/activerecord-jdbc-adapter/issues/710 (https://github.com/jruby/activerecord-jdbc-adapter/pull/746) be careful what you're using!

kendagriff:
@kares I'm not exactly sure what you're asking, but before this patch, setting either PGSSLMODE or PGREQUIRESSL did not work. After this patch, they work.

kares:
OK - now I'm really not sure whether to merge this - we turn on SSL but than disable validation by default?

jkutner:
@kares i see what you're concerned about. The `sslfactory=org.postgresql.ssl.NonValidatingFactory` was necessary in [pgjdbc](https://github.com/pgjdbc/pgjdbc) versions earlier than 9.2. From 9.2 and onward, you can use `sslmode=require`, which attempts to validate the server certificate against the CA if it exists. But if it does not exist, it still uses ssl. This configuration is required on many cloud platforms, such as Heroku.

jkutner:
@kares also, given that the default when no properties are set is to not use SSL at all, using SSL with a NonValidatingFactory is still a step up.

kares:
part of AR-JDBC 1.3.22, thanks everyone!

PR #766, opened 2016-11-28
update (official) mysql jdbc driver to 5.1.40

jayjlawrence:
@kares I hope I did this right! Submitting 5.1.40 JARs against 1.3-stable


kares:
thanks, this confirmed tests are fine. we'll however prefer the merge on master.

PR #767, opened 2016-11-30
Add views to mysql adapter

dejaner:
This adds views to mysql adapter, just like sqlite and postgres


enebo:
@dejaner great!  For any future PRs you do not need to version guard on rails-5 as we will just support 5 and up.

PR #768, opened 2016-12-10
Stop using alias_method_chain

kirs:
`alias_method_chain` has recently been removed from Rails (https://github.com/rails/rails/commit/7c848e6dd493ff236d33a0410a92f4c3e5cc3c7f).

ActiveRecord `master` will now raise because `NoMethodError`.

@enebo @kares


kares:
thanks, I believe this is done on the rails-5 branch already. it's not worth merging pieces of AR5 to master ...

PR #770, opened 2017-01-06
prevent multiple pushes of the adapter file onto $LOADED_FEATURES

abrandoned:
While the comment states that this should only be required once; if a custom resolver is used (in our case for multi-schema multi-tenancy) then this file can be loaded many times (in 1 case we have a heap dump with 4 million copies of "active_record/connection_adapters/postgresql_adapter.rb" pushed on the $LOADED_FEATURES array)

We have patched internally, but hope that something similar to this can be merged into mainline to provide the protection for anyone that may use a custom Resolver

### DIFF
--- a/lib/arjdbc/jdbc/adapter_require.rb
+++ b/lib/arjdbc/jdbc/adapter_require.rb
@@ -1,5 +1,3 @@
-require "set"
-
 module ActiveRecord
 
   if defined? ConnectionAdapters::ConnectionSpecification::Resolver # 4.0
@@ -9,7 +7,6 @@ module ActiveRecord
   else class << Base; self; end # 2.3, 3.0, 3.1 :
     # def self.establish_connection ... on ActiveRecord::Base
   end.class_eval do
-    $__ARJDBC_LOADED_FEATURES = ::Set.new
 
     # @private
     def require(path)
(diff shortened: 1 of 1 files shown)


kares:
hmm, I thought `$LOADED_FEATURES` is already treated as a Set in JRuby.

abrandoned:
I thought the same, but the patch worked and it seems to be an array in MRI and JRuby (irb output below)

JRUBY:
```ruby
jruby-9.0.1.0 :002 > $LOADED_FEATURES.class
 => Array 
jruby-9.0.1.0 :003 > 10.times do
jruby-9.0.1.0 :004 >     $LOADED_FEATURES << "DERP"
jruby-9.0.1.0 :005?>   end
 => 10 
jruby-9.0.1.0 :006 > $LOADED_FEATURES.count("DERP")
 => 10 
```

MRI:
```ruby
2.3.0 :002 > $LOADED_FEATURES.class
 => Array 
2.3.0 :003 > 10.times do
2.3.0 :004 >     $LOADED_FEATURES << "DERP"
2.3.0 :005?>   end
 => 10 
2.3.0 :006 > $LOADED_FEATURES.count("DERP")
 => 10 
```

kares:
I mean it looks like an Array but [include? does a set check](https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/util/collections/StringArraySet.java#L46) 
... no need to invent another `Set` esp. a one leaking into the system (a global)

abrandoned:
makes sense, changed to using `include?`

kares:
looking good, will cherry-pick this (eventually) for the next 1.3 release. thanks!

kares:
now squashed on 1-3-stable as: https://github.com/jruby/activerecord-jdbc-adapter/commit/767d9ed60af7d218b74e14b85370e3252b381915

PR #774, opened 2017-01-31
PostgreSQL fix RETURNING clause

### LINKED ISSUE #773: RETURNING with composite primary keys fails
I am using `activerecord-jdbcpostgresql-adapter` in one of my projects. And the problem my tests stumbled over (not a problem using CRuby with `pg`) is the following:

```
activerecord (4.2.7.1)
activerecord-jdbc-adapter (1.3.22)
activerecord-jdbcpostgresql-adapter (1.3.22)
jruby 9.1.7.0 (2.3.1) 2017-01-11 68056ae Java HotSpot(TM) 64-Bit Server VM 25.45-b02 on 1.8.0_45-b14 +jit [darwin-x86_64]
```

I insert on a join table, its primary key is a composite of two columns `collection_id` and `product_id`.

The insert it generates looks like this:

```sql
INSERT INTO "collections_products" ("product_id", "collection_id")
VALUES (1, 1)
RETURNING "collection_id,product_id"
```

_I added some newlines for readability._

If you look at the `RETURNING` clause, you can spot the error, as it is not properly converting these values to columns. And PostgreSQL throws me an `ERROR: column "collection_id,product_id" does not exist`.

I expect it to look something like this, with parantheses `(` around the two columns `)`:

```sql
INSERT INTO "collections_products" ("product_id", "collection_id")
VALUES (1, 1)
RETURNING ("collection_id", "product_id")
```

Currently I am digging in the source to find the problem, but so far no luck.

**If you know a quicker answer, please help me out.**

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/41523478-returning-with-composite-primary-keys-fails?utm_campaign=plugin&utm_content=tracker%2F136963&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F136963&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Overbryd:
> use `quote_column_names` with surrounding parantheses to construct RETURNING clause, fixes #773

Hello,

unfortunately I am not very familiar with this project. So I apologise if I am violating any guidelines for contributing. But maybe somebody can have a look at this PR and the issue I have posted to get this merged in asap.

If you point me to where I can write PostgreSQL related tests, I am happy to write a case for this issue.

Kind regards,

Lukas Rieder


kares:
looking good, test are around https://github.com/jruby/activerecord-jdbc-adapter/tree/master/test/db/postgresql ... a regression test would be great. if you decide to do so please target the 1-3-stable branch.
otherwise I will merge this as/if I circle back for a 1.3 maintenance release. in any case, thanks for looking into it!

Overbryd:
Cool, I have some time today so I will add tests to it and rebase the PR to 1-3-stable.

Overbryd:
Ok. Here is my update. I rebase the fix onto `stable-1-3`.

Then I found that the composite primary key behaviour is not default in ActiveRecord. It is enabled by a gem called `composite_primary_keys`.
Now there is an annoying hen & egg problem: Is this an issue of `composite_primary_keys`, that has no control over the underlying DB adapter, or is this an issue of the DB adapter that cannot handle composite primary keys.

I decided to implement a test case here, since this is a more or less common feature of PostgreSQL. I think the adapter should be able to handle it.

Therefore I included the correct version of `composite_primary_keys` into `gemfiles/rails42.gemfile`.

If there is a better solution, please let me know.

P.s.: It was not that straightforward to get tests to run. This helped me a lot:

```
# ant is required
brew install ant
# when working on a specific rails version, export the BUNDLE_GEMFILE
export BUNDLE_GEMFILE=gemfiles/rails42.gemfile
# set the db user
export POSTGRES_USER=postgres
# run appraisal:rails42 once (it seems to generate something that is needed for tests)
rake appraisal:rails42
# run tests
rake appraisal:rails42 test_postgres
# run a specific test
rake appraisal:rails42 test_postgres TEST=test/db/postgresql/composite_primary_keys_test.rb
```

kares:
ant? well that is just weird - maybe a relic from jruby.
this will fail on CI as the test is only supposed to run under AR 4.2, could you fix that please e.g. : 
`require 'composite_primary_keys' if ar_version('4.2')` + same condition for the test method, thanks

Overbryd:
Ok. Wrapped the whole body of the class in an `if ar_version('4.2') ... end` block. This should do the trick, everything within that TestCase depends on `AR 4.2` and `gemfiles/rails42.gemfile`.

kares:
seems like it but it must be related to your changes since only 4.2 fails, its probably due AR 4.2 upgrade.
its a bit hard to manage the dev with many adapters/drivers/versions - could you maybe revert the upgrade?

Overbryd:
Ok. I see.

As a takeaway, the upgrade to latest AR 42 seems quite straightforward. AR 42 relies on `pg_dump` to be present. But I would leave that up to you, as this is out of scope for this fix.

I amended the commit, reverting the `rails42.gemfile` to where it was before with the addition of `composite_primary_keys` for the test case.

kares:
thanks, unfortunately now we do not have the `a_custom_primary_key_test.rb` guarding against regressions
for older AR versions - bellow 4.2 ... right?

Overbryd:
Ok. I amended the commit.

I added the test for all AR versions that are supported by `composite_primary_keys` with their respective version.

kares:
excellent, thanks Lukas. I also figured out the 4.2 failure is happening even without your code. no worries.

kares:
on 1-3-stable as 8d7db029ce5e561bb98409ee6c1235d5db7f9ed9 restored the deleted test parts and made composite_primary_keys gem optional (its too AR invasive - we do not want it to interfere much) at 4ee2c02e40835d5fe6c3953262cf005ff4ed1cd0

PR #779, opened 2017-03-17
Comparing timestamps with a string causes Comparable#== deprecation warning

rsov:
calling the `to_s` on value before comparing it clears the warning

Ruby version: jruby 9.0.5.0 (2.2.3)
AR version: 4.1.15

### DIFF
--- a/lib/arjdbc/db2/column.rb
+++ b/lib/arjdbc/db2/column.rb
@@ -47,7 +47,7 @@ module ArJdbc
 
       # @override
       def type_cast(value)
-        return nil if value.nil? || value == 'NULL' || value =~ /^\s*NULL\s*$/i
+        return nil if value.nil? || value.to_s == 'NULL' || value =~ /^\s*NULL\s*$/i
         case type
         when :string    then value
         when :integer   then value.respond_to?(:to_i) ? value.to_i : (value ? 1 : 0)


kares:
@rsov thanks. could we see the warnings. its usually smelly to simply do a `to_s` due a warning.
... did you try removing it to see how it goes (as `/^\s*NULL\s*$/i` should cover the 'NULL' case) ?

rsov:
Sure, here are the  warnings, they always come as 2 per timestamp
Which gets pretty annoying when table has 4 or more timestamps
```
/home/profile/.rvm/gems/jruby-9.0.5.0@tutilities/gems/activerecord-jdbc-adapter-1.3.19/lib/arjdbc/db2/column.rb:50: warning: Comparable#== will no more rescue exceptions of #<=> in the next release.
/home/profile/.rvm/gems/jruby-9.0.5.0@tutilities/gems/activerecord-jdbc-adapter-1.3.19/lib/arjdbc/db2/column.rb:50: warning: Return nil in #<=> if the comparison is inappropriate or avoid such comparison.
```

Have not tried removing it and using just Regex

kares:
warning seems a bit weird. probably an issue in 9.0.5 (try upgrading). alternatively I've switched to `eql?`

Issue #729, opened 2016-06-03
The previously-deprecated -i switch to pg_dump is no longer supported in PostgreSQL 9.5.


evilrich:
`rake db:structure:dump` from a Rails 3 app fails with

> pg_dump: invalid option -- i

This is used in lib/arjdbc/tasks/databases3.rake line 115 (version 1.3.20).

This switch was deprecated in PostgreSQL 8.4 and was finally removed in 9.5.
##

mortonar:
#783 Should fix this(barring anything I'm missing).

kares:
resolved at https://github.com/jruby/activerecord-jdbc-adapter/pull/783

PR #783, opened 2017-04-21
Issue #729: The previously-deprecated -i switch to pg_dump is no longer supported in PostgreSQL 9.5.

mortonar:


kares:
on 1-3-stable as https://github.com/jruby/activerecord-jdbc-adapter/commit/9bc444ec903c0135fbd1362a31c0020619941467 (will be in 1.3.23) ... thanks!

PR #786, opened 2017-05-10
Error PostgreSQL::ColumnDefinition

bencarsal:
This solve the error in rails 5.1: 
Bundler::GemRequireError: There was an error while trying to load the gem 'activerecord-jdbcpostgresql-adapter'.
Gem Load Error is: uninitialized constant ActiveRecord::ConnectionAdapters::PostgreSQL::ColumnDefinition

### DIFF
--- a/lib/arjdbc/jdbc.rb
+++ b/lib/arjdbc/jdbc.rb
@@ -3,9 +3,9 @@ require 'active_support/deprecation'
 module ArJdbc
 
   # @private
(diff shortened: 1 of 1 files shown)


kares:
sorry but this breaks the intention of the code (AR40 means >= 4.0) and breaks existing (AR < 5) support
please have a look at rails-5 branch. thanks for your effort ...

PR #787, opened 2017-05-10
Error PostgreSQL::ColumnDefinition

bencarsal:
This solve the error in rails 5.1:
Bundler::GemRequireError: There was an error while trying to load the gem 'activerecord-jdbcpostgresql-adapter'.
Gem Load Error is: uninitialized constant ActiveRecord::ConnectionAdapters::PostgreSQL::ColumnDefinition

### DIFF
--- a/lib/arjdbc/jdbc.rb
+++ b/lib/arjdbc/jdbc.rb
@@ -3,9 +3,9 @@ require 'active_support/deprecation'
 module ArJdbc
 
   # @private
-  AR40 = ::ActiveRecord::VERSION::MAJOR > 3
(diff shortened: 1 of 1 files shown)


kares:
@bencarsal sorry, might have not explained this clearly but https://github.com/jruby/activerecord-jdbc-adapter/pull/786 did not mean to open the same for rails-5 branch but to actually look at how those AR constants are used - will (and already does) break elsewehere :( https://github.com/jruby/activerecord-jdbc-adapter/commit/ddb4ba55911d3fe7f786ed05ff84f5571e211bf4#commitcomment-22127885

this fixed immediate issues of yours but is breaking much more - you should actually try to run some tests.
you won't be able to provide quality fixes by editing stuff on GH :wink:

PR #789, opened 2017-05-14
[as400] Fixed ordering when limit and offset present

rsov:
Re-wrote the query in `replace_limit_offset!` method in db2 adapter. SQL is simple to follow and it handles ordering

### DIFF
--- a/lib/arjdbc/db2/adapter.rb
+++ b/lib/arjdbc/db2/adapter.rb
@@ -492,29 +492,26 @@ module ArJdbc
 
     # @private shared with {Arel::Visitors::DB2}
     def replace_limit_offset!(sql, limit, offset, orders = nil)
-      limit = limit.to_i
+      # Ordering is done somewhere before this method gets called
 
-      if offset # && limit
-        over_order_by = nil # NOTE: orders matching got reverted as it was not complete and there were no case covering it ...
+      # Create the limit and offset sql
+      param_sql = ''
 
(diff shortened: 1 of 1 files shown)


kares:
thanks, will this work in other DB2 systems besides AS400? (I mean the 'regular' DB2 one can install)
could you please verify by running the suite ?

rsov:
I tried to connect to the db2 side of our as400 but could not get it to work. But I'll give it a try once we figure out why its not working

kares:
ping ... if you're really into having this in for next 1.3 release. isn't planned yet but its been a month or so.

rsov:
@kares Sorry, I could not connect to DB2, only AS400. The tests that were originally failing are not passing but I am not sure what else it will effect

rsov:
@kares Turns out there is no way I can test this. Its a product that needs purchasing and setting up. Currently we are running JDBC on AS400, because could not get the DB2 connection to work

kares:
@rsov that's weird I had it installed just fine on a (virtual) Linux box. there should be a developer edition.
guess one needs to register though ... https://www-01.ibm.com/marketing/iwm/iwm/web/preLogin.do?source=swg-db2dtrial

rsov:
@kares I did not know about this product. Thank you. Will look into it in a few days

kares:
did a quick run, seems that tests run fine and actually 2 errors went away, thus put this in https://github.com/jruby/activerecord-jdbc-adapter/commit/15d7d1f60c80662e5d237503780c0621ae698224

PR #791, opened 2017-05-14
[as400] Database tasks fix

rsov:
AS400 was missing the `DB2DatabaseTasks` module


kares:
excellent, a release went out just yesterday, but you can hopefully use `:git => ...` in the mean-time

PR #792, opened 2017-05-14
[as400] Don't throw an exception if sequence number was defined

rsov:
AS400 throws an exception if you try to run `ADDRPYLE SEQNBR(9876) MSGID(CPA32B2) RPY('I')` more than once per login. This was causing some issues in the testing environment


kares:
btw. did you try https://github.com/pierrickrouxel/activerecord-jdbcas400-adapter ... supposed to be better than AR-JDBC. the idea was to move AS400 out of here - since we can not test it, thus it will always come last (not that anyone tests other DB2 systems much but I have tried and there's always a possibility to install the database compared to AS/400)

rsov:
@kares Yes i did. It's a good idea to put all the as400 hacks into a separate gem. I could not get the AR 4.2 tests to run with it. They have a `rake test:db2` command which runs the tests with db2 driver but nothing for the as400. Have to patch in something so it picks up on the as400 one

kares:
@rsov OK - was just worried that the as400 adapter might be a better suit for you.
no worries if you have tried it already.

rsov:
@kares Thanks man! Will probably look into how to connect to the db2 side of the as400

Issue #713, opened 2016-02-26
[mssql] to_sql with limit not working


soemo:
Hi,

I'm running a rails app with
JRuby 9.0.5.0
Rails 4.2.5.1
MSSQL Server 2008

In activerecord-jdbc-adapter 1.3.19 and on master I get an error, when i use to_sql on a limit query.
````ruby
User.where(id:[1,2]).limit(1).to_sql

NoMethodError: undefined method `map' for nil:NilClass
from /home/soeren/.rbenv/versions/jruby-9.0.5.0/lib/ruby/gems/shared/gems/arel-6.0.3/lib/arel/collectors/bind.rb:22:in `substitute_binds'
from /home/soeren/.rbenv/versions/jruby-9.0.5.0/lib/ruby/gems/shared/gems/activerecord-4.2.5.1/lib/active_record/relation.rb:552:in `to_sql'
from (irb):2:in `<eval>'
from org/jruby/RubyKernel.java:976:in `eval'
from org/jruby/RubyKernel.java:1287:in `loop'
from org/jruby/RubyKernel.java:1096:in `catch'
from org/jruby/RubyKernel.java:1096:in `catch'
from /home/soeren/.rbenv/versions/jruby-9.0.5.0/lib/ruby/gems/shared/gems/railties-4.2.5.1/lib/rails/commands/console.rb:110:in `start'
from /home/soeren/.rbenv/versions/jruby-9.0.5.0/lib/ruby/gems/shared/gems/railties-4.2.5.1/lib/rails/commands/console.rb:9:in `start'
from /home/soeren/.rbenv/versions/jruby-9.0.5.0/lib/ruby/gems/shared/gems/railties-4.2.5.1/lib/rails/commands/commands_tasks.rb:68:in `console'
from /home/soeren/.rbenv/versions/jruby-9.0.5.0/lib/ruby/gems/shared/gems/railties-4.2.5.1/lib/rails/commands/commands_tasks.rb:39:in `run_command!'
from /home/soeren/.rbenv/versions/jruby-9.0.5.0/lib/ruby/gems/shared/gems/railties-4.2.5.1/lib/rails/commands.rb:17:in `<top>'
from org/jruby/RubyKernel.java:937:in `require'
````

Its works correct without limit 
````ruby
User.where(id:[1,2]).to_sql
````

Any ideas?

rob99:
I found that if I change 

[...]

PR #794, opened 2017-05-15
Merged with rails-sqlserver/activerecord-sqlserver-adapter

keeguon:
I merged the Arel visitors with rails-sqlserver/activerecord-sqlserver-adapter, it fixes some of the errors with LIMIT (see #713).


kares:
there was a reason why the original code was not used in the end. namely that it caused regressions with the suite. could you please run the suite before and after the change for comparison what failures there are ?

kares:
closing since there was no resolution whether it regressed or not, recall it previously did break (last time I tried to bring the sqlserver adapter pieces over).
... not looking into MS-SQL myself since I simply lack the motivation to do that much house-keeping, sorry.
hopefully, someday there will be resources to get support for MS-SQL right, until than no risky changes ...

Issue #725, opened 2016-05-13
[db2] serialize with custom class does not work


rsov:
Active Record - 4.2
ArJdbc - 1.3.20
JRuby - 9.1.0.0
Database - DB2/As400

This worked on AR-3.2

Ran into an issue where return value from `CustomDateSerialize#dump(value)` does not get written to database

Purpose of it is to turn a Date class into an integer (YYYYMMDD)

``` ruby
class CustomDateSerialize

  def self.dump(value)
    puts 'this shows up, so the method does gets called'

    if value.respond_to? 'strftime'
      value.strftime('%Y%0m%0d').to_i
    else
      value
    end
  end

  def self.load(value)
    Date.strptime(value.to_s, '%Y%m%d') rescue nil
  end

end
```

Expected result: 
`Date.new(2016, 1, 1) -> 20160101`

Gotten:
`Date.new(2016, 1, 1) -> 2016-01-01` (Default date.to_s)
##

kares:
yes - unfortunately DB2 adapter is hard to support ... esp. AS400 (do not have access to such machine). 
AR 4.2 integration simply needs more work. there's also an attempt to get it out [into a standalone gem](https://github.com/pierrickrouxel/activerecord-jdbcas400-adapter).

rsov:
That makes sense. We unfortunately have an as400. How is the AR 5 integration for it looking? 

If we figure it out, will submit a pull request

kares:
there's some progress on a branch but needs more work.
not sure if someone will popup or sponsor the Rails 5 integration. we're somehow quite low on help on this front.

PR #795, opened 2017-05-19
[AR42][DB2] Fixed issue with serialization not happening

rsov:
This is the fix for #725 bug

Added tests to confirm the behaviour

For some reason the the DB2 adapter's `quote` method is not getting value after its been serialized. So we check the column is  `ActiveRecord::Type::Serialized` and then cast it for database

### DIFF
--- a/lib/arjdbc/db2/adapter.rb
+++ b/lib/arjdbc/db2/adapter.rb
@@ -363,6 +363,11 @@ module ArJdbc
         if column.respond_to?(:primary) && column.primary && column.klass != String
           return value.to_i.to_s
         end
+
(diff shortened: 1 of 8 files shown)


kares:
thanks @rsov ... as you probably already noticed your change effected other adapters and caused failures.
you can easily reproduce with SQLite using AR 3.0 (know its old but the stable branch should not break) ...
test simply needs updating (or isolating) https://travis-ci.org/jruby/activerecord-jdbc-adapter/jobs/234553859

rsov:
@kares Yes, I've noticed it but didn't know what to do. Isolating the test is probably a good idea as it effects the db2 adapter only

PR #797, opened 2017-06-21
Fixed OID type tests and removed conditional tests

rdubya:
- cast_type was removed from the Column class in ActiveRecord in favor of retrieving types from the model itself
-- https://github.com/rails/rails/commit/155b1b7fe3a1d231fb98a6fb04a21f6eb190b98f
-- https://github.com/rails/rails/commit/158c7eb1d61a28452e0aafd1e05314352eea2749

- accessor was removed from the Column class in ActiveRecord
-- https://github.com/rails/rails/commit/2eb547a4d9cfb4ed1c92397921402048162f5fc8
-- https://github.com/rails/rails/commit/b93b39eff6829ee05ffec1cc8c505f69cbb53fdc#diff-2524057922ad16b58e6b18b2f35b5de2

- the ActiveRecord::ConnectionAdapters::PostgreSQLAdapter::OID::Integer class was removed and now uses ActiveRecord::Type::Integer

### DIFF
--- a/test/db/postgresql/oid_types_test.rb
+++ b/test/db/postgresql/oid_types_test.rb
@@ -6,12 +6,6 @@ class PostgresqlOOIDTypesTest < Test::Unit::TestCase
 
   OID = ActiveRecord::ConnectionAdapters::PostgreSQLAdapter::OID
 
-  #OID::TypeMap.class_eval do
-  #  def size
-  #    @mapping.size
-  #  end
-  #end
-
   def setup
     @connection = ActiveRecord::Base.connection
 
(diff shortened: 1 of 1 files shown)


rdubya:
I removed the conditionals checking for JRUBY_VERSION as well. Let me know if we need to keep those. I was under the impression that Rails 5 wouldn't run on any JRuby versions that wouldn't have that defined but wasn't sure if there was some other reason for the check.

test/db/postgresql/oid_types_test.rb
```diff
@@ -40,31 +34,21 @@ class SomeSample < ActiveRecord::Base
...
-      column = adapter.columns('some_samples').find { |c| c.name.to_sym == :int }
-    end
+    column = adapter.column_for('some_samples', :int)
     assert_not_nil column
-    assert_instance_of OID::Integer, oid_type(column)
-  end if ar_version('4.1')
+    assert_instance_of ActiveRecord::Type::Integer, column.type
+  end
 
   def test_returns_column_accessor_for_hstore
```
rdubya: I'm not sure if this test is even needed anymore. Now it is basically just testing that the OID::Hstore class has it defined which probably belongs in another group of tests if it isn't already covered elsewhere.

enebo: @rdubya yeah we seem to have lots of perhaps too low-level tests as well which can probably go away.  I think an older motivation was to add a mixture of common tests from AR into our suite to not have to run full AR suite (or perhaps merely for regressions -- I don't actually know).  So we can start deleting stuff when it is easily verified running against Rails itself.

enebo:
@rdubya yeah this will be 9k and above only and the plan we want to pursure is not supporting all versions of all things but make new numbered releases to coincide with rails releases.  I am hoping Ruby is stable enough where JRuby specific versions will no longer be a concern (although never say never). 

For now assume JRuby 9k+.  This means pretty much all version checks can go away (I am really hoping Ruby 2.4 adds no wrinkles).

test/db/postgresql/oid_types_test.rb
rdubya: @enebo Sound good. I have another PR about ready with some more fixes so I'll remove it in that one.

PR #800, opened 2017-07-07
Fix postgres array support for Rails 5.0

rdubya:
I tried to reuse as much of the built in Rails logic as possible since we don't need to worry about being backwards compatible anymore. It slimmed down the amount of code needed for the adapter and actually uncovered some places where the jdbc postgres adapter doesn't support the same options as the built-in postgres adapter. There is still a few more methods that can probably be removed but they don't have as close to the same logic.


lib/arjdbc/postgresql/base/array_encoder.rb
```diff
@@ -0,0 +1,21 @@
+# This implements a basic encoder to work around ActiveRecord's dependence on the pg gem
+module PG
```
rdubya: I'm not sure if this is the best way to do the array encoding or not. I couldn't figure out how it was previously being done, so if you have info on that and would like me to try to go that route I can dig into it. The one difference with this way that I noticed is that all of the values are always quoted inside the array, even if they are numbers. It doesn't seem like that matters based on the tests but it is something to keep in mind.

lib/arjdbc/postgresql/base/array_encoder.rb
```diff
@@ -0,0 +1,21 @@
...
+  module TextEncoder
+    class Array
+
+      def initialize(name:, delimiter:)
+        @type = if name == 'string[]'.freeze
```
rdubya: I called out string explicitly because it currently get's converted to 'character varying' which the driver didn't seem to be able to handle.

lib/arjdbc/postgresql/adapter.rb
```diff
@@ -602,7 +511,11 @@ def query(sql, name = nil)
       log(sql, name) do
         result = []
         @connection.execute_query_raw(sql, nil) do |*values|
-          result << values
+          # We need to use #deep_dup here because it appears that
```
rdubya: I wasn't sure the best way to fix this. The issue was that the result array would end up with duplicates of the last item for all of its entries. My theory on why that happens is that the same Java object is being used for the callback and just having its values replaced. It then gets converted to be a Ruby object, but it keeps a reference to the same memory instead of creating a new copy of the object. I think this is the related code:

```
   private IRubyObject yieldResultRows(final ThreadContext context, final Ruby runtime,
            final Connection connection, final ResultSet resultSet,
            final Block block) throws SQLException {

        final ColumnData[] columns = extractColumns(runtime, connection, resultSet, false);

        final IRubyObject[] blockArgs = new IRubyObject[columns.length];
        while ( resultSet.next() ) {
            for ( int i = 0; i < columns.length; i++ ) {
                final ColumnData column = columns[i];
                blockArgs[i] = jdbcToRuby(context, runtime, column.index, column.type, resultSet);
            }
            block.call( context, blockArgs );
        }

        return runtime.getNil(); // yielded result rows
    }
```

Notice that the same object is being used as the arguments object when calling the block. I didn't have a good way of verifying my theory though. I think it could be fixed by creating a new object each time, but I wasn't sure where it was better to fix it. If you think its better to do it on the Java side, then I can try to bite that off in another PR

lib/arjdbc/postgresql/base/array_encoder.rb
enebo: I don't know how it was being done but perhaps @kares knows.

enebo: I think forward progress here and in your other comments is probably a bit more important than stalling to do this differently.  If @kares has any insight we will just follow up on it.

lib/arjdbc/postgresql/base/array_encoder.rb
enebo: Seems reasonable as a workaround.  My only other thought is perhaps the version of JDBC somehow fixes something?

PR #801, opened 2017-07-07
Fix postgres schema dumping for Rails 5.0

rdubya:
I was able to mostly reuse the built in functionality from ActiveRecord's Postgres adapter. There were a couple of small output changes in ActiveRecord for indexes and bigints that required tests to be updated.


lib/arjdbc/postgresql/adapter.rb
```diff
@@ -866,6 +852,11 @@ def column_definitions(table_name)
            AND a.attnum > 0 AND NOT a.attisdropped
          ORDER BY a.attnum
       end_sql
+
+      # Force the notnull attribute to a boolean
+      rows.each do |row|
+        row[3] = row[3] == 't' if row[3].is_a?(String)
```
rdubya: This previously happened in the #columns method but doing it here allows us to let the ActiveRecord adapter handle the bulk of the work.

PR #804, opened 2017-07-18
Postgres rails 5.0.4 fixes

rdubya:
In ActiveRecord 5.0.3 or 5.0.4 they updated `PGConn` references to be `PG::Connection` since `PGConn` was deprecated.

Should I also include the compiled jar file for the adapter with the pr or should that not be in the repo?

### DIFF
--- a/lib/arjdbc/postgresql/base/pgconn.rb
+++ b/lib/arjdbc/postgresql/base/pgconn.rb
@@ -1,6 +1,5 @@
 module PG
-  class Connection # emulate PGconn#unescape_bytea due #652
-    # NOTE: on pg gem ... PGconn = (class) PG::Connection
+  class Connection # emulate PG::Connection#unescape_bytea due #652
     def self.unescape_bytea(escaped)
       ArJdbc::PostgreSQL.unescape_bytea(escaped)
     end
--- a/lib/arjdbc/postgresql/base/pgconn.rb
+++ b/lib/arjdbc/postgresql/base/pgconn.rb
@@ -1,5 +1,5 @@
-module ActiveRecord::ConnectionAdapters::PostgreSQL::OID
-  class PGconn # emulate PGconn#unescape_bytea due #652
+module PG
+  class Connection # emulate PGconn#unescape_bytea due #652
     # NOTE: on pg gem ... PGconn = (class) PG::Connection
     def self.unescape_bytea(escaped)
       ArJdbc::PostgreSQL.unescape_bytea(escaped)
--- a/src/java/arjdbc/jdbc/RubyJdbcConnection.java
+++ b/src/java/arjdbc/jdbc/RubyJdbcConnection.java
@@ -1791,7 +1791,7 @@ public class RubyJdbcConnection extends RubyObject {
     protected static IRubyObject typeCastFromDatabase(final ThreadContext context,
         final IRubyObject adapter, final RubySymbol typeName, final RubyString value) {
         final IRubyObject type = adapter.callMethod(context, "lookup_cast_type", typeName);
-        return type.callMethod(context, "type_cast_from_database", value);
+        return type.callMethod(context, "deserialize", value);
     }
 
     protected IRubyObject dateToRuby(final ThreadContext context,
--- a/lib/arjdbc/postgresql/adapter.rb
+++ b/lib/arjdbc/postgresql/adapter.rb
@@ -973,9 +973,11 @@ module ArJdbc
     def translate_exception(exception, message)
       case exception.message
       when /duplicate key value violates unique constraint/
-        ::ActiveRecord::RecordNotUnique.new(message, exception)
+        ::ActiveRecord::RecordNotUnique.new(message)
       when /violates foreign key constraint/
-        ::ActiveRecord::InvalidForeignKey.new(message, exception)
+        ::ActiveRecord::InvalidForeignKey.new(message)
+      when /value too long/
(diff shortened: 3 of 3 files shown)


src/java/arjdbc/jdbc/RubyJdbcConnection.java
```diff
@@ -1791,7 +1791,7 @@ public static IRubyObject setRawDateTime(final IRubyObject self, final IRubyObje
     protected static IRubyObject typeCastFromDatabase(final ThreadContext context,
         final IRubyObject adapter, final RubySymbol typeName, final RubyString value) {
         final IRubyObject type = adapter.callMethod(context, "lookup_cast_type", typeName);
-        return type.callMethod(context, "type_cast_from_database", value);
+        return type.callMethod(context, "deserialize", value);
```
enebo: @rdubya I assume this is a general Rails 5 change?  My msqlite changes did not hit this (from what I recall) so I don't know if this is common for other adapters but it is in RubyJdbcconnection.java.

rdubya: Yeah, ActiveRecord now uses `deserialize` on the type and removed the `type_cast_from_database` method. It seems like this method is only called in certain cases, so I can see why you potentially wouldn't have seen it. There were only a handful of tests that were broken because of it in the postgresql adapter.

enebo: @rdubya ah great.  Thanks for the explanation.

lib/arjdbc/postgresql/base/pgconn.rb
```diff
@@ -1,8 +1,8 @@
-module ActiveRecord::ConnectionAdapters::PostgreSQL::OID
-  class PGconn # emulate PGconn#unescape_bytea due #652
```
kares: the idea here was not to pollute yet another name-space - AR-JDBC managed to keep it name-spaced under its own.
probably not that relevant but a `defined? PG` can now yield surprises with AR-JDBC despite not having any PG loaded

enebo: @kares Is your concern that some other gem will check PG and think native gem is loaded instead?

rdubya: I can hopefully make a pull request tomorrow to try to keep them scoped to where they are accessed if that would be better. Otherwise, we have to override other logic so that we can bypass the references. The ideal solution would probably be to get some patches into ActiveRecord to make it less PG dependent, but it seems like that would be better to handle when working on 5.1 or 5.2 compatibility, if the Rails team would accept those changes.

enebo: @rdubya so I assumed we used PG because activerecord methods are scoped using some methods from PG.  If so then I would rather leverage the difficulties of that then clone all of the methods which call PG ones in AR.  

I agree the correct long term solution may be to abstract/remove PG in a way which we do not need to stomp on the same constant and "pretend".

kares: yeah its a lack of proper abstraction on AR's end ... 
have seen driver 'feature' detection in gems that is why I recommend to keep not using a name-space known to be used with another gem if we're not loading up that gem at runtime.

enebo: @kares I know it sounds like a lot of work but we have had some level of commitment from people working on Rails to help make our integration less painful.  So perhaps we can at least scope out how AR could do things in a way where we don't emulate/fake-out PG and then get them to commit to our plan?

Also do you happen to know which gems offhand?  I am curious if perhaps we can just PR their checks in a reasonable way where we don't trip over them.

  My main desire is to try and reduce the amount of code within ARJDBC to minimize how much we need to update in the future.  In the case of sqlite3 I think we can eliminate 95% of the Ruby we have with a couple of changes (on rails5).  For postgresql it looks much less possible but I guess I am hoping for some upstream AR changes to help eventually reduce that.

My real desire is I wish Rails AR would just use sequel for all adapters.  We would not even need to have this project :|

kares: @enebo is PG all over the place in 5.0 at AR's, wasnt that hard to support over the years :)?
... you could still use the trick of not polluting by putting PG under a common name-space
e.g. `ActiveRecord::ConnectionAdapters::PostgreSQLAdapter::PG` have used similar around here

I do not have anything off hand, its been a while - in general AR-JDBC has code in at least 3 top level name-spaces, so this 4-th seemed like it can be avoided ... but I have no strong objections if its to be kept. unless I get "hired" for AR-JDBC tech.debt elimination :)

enebo: @kares I agree putting PG as non-global is best if it works.  I respect what you say but I also think we need to work with the idea we can change things upstream (AR) to make things easier since there clearly is not enough contributions to keep this up to date.  So I am probably too optimistic to thing we can affect some change and reduce or eliminate this?

rdubya: @kares Just a heads up that https://github.com/jruby/activerecord-jdbc-adapter/pull/809 makes the change to keep `PG` scoped to where it is referenced.

I'm currently working on making it so that AR-JDBC reuses as much of the core code as possible, but there is definitely more that could be reused if some more methods can be moved around in ActiveRecord itself.

PR #805, opened 2017-07-19
Solution error: Gem Load ColumnDefinition for version 5.1

ybart:
The propose is to solution this error: Gem Load Error is: uninitialized constant ActiveRecord::ConnectionAdapters::PostgreSQL::ColumnDefinition

This allowed to pass `uninitialized constant ActiveRecord::ConnectionAdapters::PostgreSQL::ColumnDefinition` error on Rails 5.1. However, some more work seems required in my case. I am now experiencing the following issue using this patch :

```
NoMethodError (undefined method `cast' for "character varying":String):

vendor/cache/activerecord-jdbc-adapter-6dbb2441c0db/lib/arjdbc/jdbc/column.rb:30:in `initialize'
```

### DIFF
--- a/lib/arjdbc/jdbc.rb
+++ b/lib/arjdbc/jdbc.rb
@@ -3,9 +3,9 @@ require 'active_support/deprecation'
 module ArJdbc
 
   # @private
-  AR40 = ::ActiveRecord::VERSION::MAJOR > 3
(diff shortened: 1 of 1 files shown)


kares:
... this isn't the way to go (need to touch the code :)), those constants were meant to return true for any >=

ybart:
I realized there was already a discussion about this commit explaining why this did not solve the problem and was only replacing the error message with some other, so I closed :)

I investigated a bit, but the fix does not seem to be trivial.

PR #809, opened 2017-07-20
Scope PG namespace

rdubya:
This makes it so that the PG modules/classes that we are emulating will be scoped to where they are referenced so that it doesn't look like the pg gem is loaded for others that may be checking `defined? PG`. This addresses concerns raised in: https://github.com/jruby/activerecord-jdbc-adapter/pull/804#discussion_r128342423


kares:
@rdubya very nice, thank you so much for all of your efforts ... :purple_heart: it!

PR #810, opened 2017-07-21
Postgres adapter cleanup

rdubya:
Updated the adapter to use more of ActiveRecord's logic. Most of the changes in here are just deleting methods that are now coming from `ActiveRecord::ConnectionAdapters::PostgreSQL::DatabaseStatements`

### DIFF
--- a/lib/arjdbc/postgresql/adapter.rb
+++ b/lib/arjdbc/postgresql/adapter.rb
@@ -3,6 +3,7 @@ ArJdbc.load_java_part :PostgreSQL
 
 require 'ipaddr'
 require 'active_record/connection_adapters/postgresql/column'
+require 'active_record/connection_adapters/postgresql/database_statements'
 require 'active_record/connection_adapters/postgresql/quoting'
 require 'active_record/connection_adapters/postgresql/schema_dumper'
 require 'active_record/connection_adapters/postgresql/schema_statements'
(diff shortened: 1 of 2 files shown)


lib/arjdbc/postgresql/adapter.rb
```diff
@@ -1012,10 +774,39 @@ module ActiveRecord::ConnectionAdapters
...
+                     'release_savepoint', 'current_savepoint_name']
+    SAVED_METHOD_PREFIX = 'jdbc_adapter'
+
+    SAVED_METHODS.each do |base_name|
+      alias_method :"#{SAVED_METHOD_PREFIX}_#{base_name}", base_name
+    end
+
+    # Keep this one separate because we don't want it to be automatically redefined
+    alias_method :"#{SAVED_METHOD_PREFIX}_exec_insert", :exec_insert
+
```
enebo: I am confused by this logic...Can you include DatebaseStatements above the definitions you want to save?

rdubya: @enebo The functionality for all of these is defined in the `JdbcAdapter` then it gets overwritten by the stuff from `DatabaseStatements` when it is included. Unfortunately, all of these methods have `PG` specific functionality tied to them, so we need to keep the functionality that was defined in `JdbcAdapter`. This was a quick and dirty way to do that without trying pull them all into a module and then have the same module included in the `JdbcAdapter` and again here in its subclass. I think in the future, we'll be able to use a similar setup to how the sqlite adapter works where it no longer subclasses `JdbcAdapter` and at that point it will make sense to move this stuff into a module or multiple modules, once we have a better idea of what all is going to need to be handled this way. I can try to bite that off now but thought it was better to keep this PR smaller and do that in a  future PR.

PR #812, opened 2017-07-26
Fix explain support tests

rdubya:
This fixes a failure that happens in the explain support test for binds when not using prepared statements. If prepared statements is enabled it would pass.

I also removed the Postgres adapter's explain logic in favor of using the explain logic from ActiveRecord's Postgres adapter.


lib/arjdbc/postgresql/adapter.rb
```diff
@@ -218,15 +218,23 @@ def standard_conforming_strings?
...
-    def supports_migrations?
-      true
-    end
+    def supports_ddl_transactions?; true end
 
-    # Does PostgreSQL support finding primary key on non-Active Record tables?
-    def supports_primary_key?
-      true
-    end
+    def supports_explain?; true end
```
rdubya: This was moved here from the explain_support.rb file. The others were just moved around so that I could group them into once place if they are just returning `true`.

PR #813, opened 2017-07-28
Fixes for Postgres Adapter against ActiveRecord test suite

rdubya:
ActiveRecord limited the minitest version used and I had to update mocha to the latest because of some issues with JRuby compatibility. Also included a couple of small fixes that cleaned up a few of the tests

### DIFF
--- a/lib/arjdbc/jdbc/adapter.rb
+++ b/lib/arjdbc/jdbc/adapter.rb
@@ -289,12 +289,14 @@ module ActiveRecord
 
       # @override
       def reconnect!
+        super
         @connection.reconnect! # handles adapter.configure_connection
         @connection
       end
 
       # @override
       def disconnect!
+        super
         @connection.disconnect!
       end
 
--- a/Gemfile
+++ b/Gemfile
@@ -25,9 +25,10 @@ gem 'appraisal', '~> 0.5.2', :require => nil
 
 # appraisal ignores group block declarations :
 
+gem 'minitest', '< 5.3.4'
 gem 'test-unit', '~> 2.5.4', :group => :test
 gem 'test-unit-context', '>= 0.4.0', :group => :test
-gem 'mocha', '~> 0.13.1', :require => nil, :group => :test
+gem 'mocha', '~> 1.2', :require => nil, :group => :test
 
 gem 'simplecov', :require => nil, :group => :test
 gem 'bcrypt-ruby', '~> 3.0.0', :require => nil, :group => :test
--- a/lib/arjdbc/postgresql/adapter.rb
+++ b/lib/arjdbc/postgresql/adapter.rb
@@ -530,18 +530,6 @@ module ArJdbc
       end
     end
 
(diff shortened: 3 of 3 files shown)


lib/arjdbc/jdbc/adapter.rb
```diff
@@ -289,12 +289,14 @@ def active?
 
       # @override
       def reconnect!
+        super
```
kares: so this actually helped a piece of test passing?
... not sure if it plays nicely with JNDI - they used to have some weird assumptions (for mostly embed adapters)

rdubya: Yeah, they must have cleaned it up or something. Here is the relevant code from the abstract adapter:

```
      # Disconnects from the database if already connected, and establishes a
      # new connection with the database. Implementors should call super if they
      # override the default implementation.
      def reconnect!
        clear_cache!
        reset_transaction
      end

      # Disconnects from the database if already connected. Otherwise, this
      # method does nothing.
      def disconnect!
        clear_cache!
        reset_transaction
      end

      ###
      # Clear any caching the database adapter may be doing, for example
      # clearing the prepared statement cache. This is database specific.
      def clear_cache!
        # this should be overridden by concrete adapters
      end

      def reset_transaction #:nodoc:
        @transaction_manager = ConnectionAdapters::TransactionManager.new(self)
      end
```

The issue was that the transaction state wasn't being updated, so the adapter would think it was still inside an active transaction even though the connection had been closed.

kares: ah ok, that makes sense - should review if this isn't relevant in current stable as well, thanks

rdubya: It looks like ActiveRecord 4.2 has roughly the same code, so I would guess that the same issue exists in the stable release. My current codebase isn't in a state that I can easily test the stable branch, but I can dig into it if you think its something that should be fixed.

kares: yeah it totally makes sense on first sight - wonder why it doesn't matter much (no issues reported).
no need to if you're busy with 5.x, maybe could you point me to the tests that were failing before?

rdubya: The tests that were failing are:

```
  1) Failure:
ActiveRecord::AdapterTestWithoutTransaction#test_transaction_state_is_reset_after_a_reconnect [~/repos/rails/activerecord/test/cases/adapter_test.rb:324]:
Failed assertion, no message given.


  2) Failure:
ActiveRecord::AdapterTestWithoutTransaction#test_transaction_state_is_reset_after_a_disconnect [~/repos/rails/activerecord/test/cases/adapter_test.rb:331]:
Failed assertion, no message given.
```

kares: thanks, I am starting to recall this worked differently for some reason/purpose.
anyway its probably worth a new investigation how/when it changed. 
... although seems less important by now :) thanks again!

PR #815, opened 2017-09-07
Initial Rails 5.0 prepared statement support

rdubya:
This updates the adapter to handle the newer ActiveRecord::Attribute bindings that are used in Rails 5+. There are still some edge cases that need to be worked through but the postgres and sqlite test suites have close to the same passing rate for when prepared statements are enabled or disabled now.

- I updated the postgres jdbc driver to match what is currently in the master branch, it is still a few versions behind however
- I updated the source/target for compilation to Java 7 to support having switch statements with Strings. I can switch it back and remove the switch statements if you think we need to support Java 6 still. It was EOL several years ago though.
- There is currently a little weirdness in the jdbc/adapter.rb file because of the transition of the postgres adapter to no longer be a subclass of it. I'm not sure what the best way to handle that was. If somebody starts working on one of the other adapters, they'll just need to uncomment those lines.

Future work related to this PR
- Fix large object support for databases that make separate calls for them, I just commented it out for now since it doesn't apply to the postgres or sqlite adapters, but it shouldn't be too hard to update the calls to pass ActiveRecord::Attribute objects and work with the new prepared statement code
- Fix some edge cases that are still broken
- Continue working on making the Postgres adapter directly subclass ActiveRecord::ConnectionAdapters::AbstractAdapter


rdubya:
@kares @enebo Can either of you take a look at this sometime this week/weekend? I'll probably try to start working on more of the future work I mentioned in the description next week but I want to make sure the base is solid before I spend too much more time on it.

kares:
hey @rdubya ... have tried looking twice already - just forgot to give feedback :

* don't mind the Java 7 switch but users might expect it to work on Java 6?
  pbly not since there wasn't but one pre release with 5.x

* the prepared statement support seems reasonable - unfortunately one can only tell by looking into AR internals if its done right :)

* the Postgres jdbc adapter change seems unfortunate to be done here

the JDBC driver wasn't upgraded (supported) since 9.4 introduced some (internal) API changes we're using, there's [an attempt to deal with the change in a backwards compatible way but somehow failed to deliver a workable version for quite some time](https://github.com/jruby/activerecord-jdbc-adapter/pull/746)

rdubya:
@kares Thanks for the feedback.

- On the Java 7 switch: I'll keep it in there and if it becomes an issue during pre-release it wouldn't be too hard to revert, just helps keep the code a little cleaner and potentially a perf boost

- The jdbc adapter update is just updating it to the version that is currently being used by the 1.3 version of this gem. The branch was split so long ago that it was still using postgres version 9.3. It seems like at this point that there is going to be enough different in this Rails 5 stuff that we won't be able to be backwards compatible anymore (without a bunch of work), so it may be better to just update to the latest postgres driver as part of this branch and call it a breaking change. I would do that as a separate PR though.

kares:
> The jdbc adapter update is just updating it to the version that is currently being used by the 1.3 version of this gem. The branch was split so long ago that it was still using postgres version 9.3. It seems like at this point that there is going to be enough different in this Rails 5 stuff that we won't be able to be backwards compatible anymore (without a bunch of work), so it may be better to just update to the latest postgres driver as part of this branch and call it a breaking change. I would do that as a separate PR though.

that's all well and fine but we're not forward compatible - all of the current 1.3 adapters gem versions would break once a 9.4 driver gem version is pushed. so it better get managed out right - I understand its much more work and we're quite very low on resources but that would be the way to do it properly. 
so I am raising it as a concern ... going to limit next 1.3 release to < 9.4 version, so we can at least hope most (old - in terms of long-term "legacy") users will be on latest 1.3.x adapter release.

rdubya:
Ah, I didn't realize there were issues with the 9.4 version since that is what we are currently using in production. I don't think it will be an issue to pull the changes out of here though and we can handle it as a separate PR.

kares:
well the driver probably only breaks bad (ClassNotFoundException) if you're using some of the custom types.
will look into finishing up for #710 (#746)

on Java 7 ... I realized that AR 5.x needs JRuby 9K at minimum which depends on Java >= 7 thus its all good - thanks for the reminder and the update.

rdubya:
@kares If you want to take another look at this, I think it should be in a state that is worth merging in and closing this PR. The changes that I made in the last commit may actually fix the issue with the 9.4 driver since it no longer tries to use the `Jdbc4Array` class.

There is still 1 test failing when saving binary data that I'm not sure if it is an issue with the test or where in the process it is breaking down, but I think it is more beneficial at this point to move on to some of the larger issues that still exist with the adapter.

rdubya:
@kares I think I've cleaned up pretty much all of the sqlite and postgres specs that were failing due to the prepared statement changes I made. I think the only one left is the large binary data test, which I'm not sure where it is losing the data since it seems to have everything up to the point of being passed into the prepared statement. That can probably be addressed in a separate PR unless you think it needs to be addressed along with these changes.

Is anything else that you think I should change? I have another PR waiting that was built on top of this branch that fixes some additional test failures that are not prepared statement related, but I want to make sure this gets in first.

lib/arjdbc/abstract/database_statements.rb
```diff
@@ -5,63 +5,25 @@ module Abstract
...
-          log(sql, name || 'SQL', binds) { @connection.execute_delete(sql, binds) }
+      # It appears that at this point (AR 5.0) "prepare" should only ever be true
+      # if prepared statements are enabled
+      def exec_query(sql, name = nil, binds = [], prepare: false)
+        if without_prepared_statement?(binds)
+          execute(sql, name)
         else
-          sql = to_sql(sql, binds) if sql.respond_to?(:to_sql)
-          log(sql, name || 'SQL') { @connection.execute_delete(sql) }
+          binds = convert_legacy_binds_to_attributes(binds) if binds.first.is_a?(Array)
```
kares: why we need to do this, is it only due some tests (maybe they're stupid tests that would not pass on MRI) ?

lib/arjdbc/common_jdbc_methods.rb
```diff
@@ -9,5 +9,19 @@ def initialize(connection, logger = nil, config = {})
...
     end
+
+    def translate_exception(e, message)
+      # we shall not translate native "Java" exceptions as they might
+      # swallow an ArJdbc / driver bug into a AR::StatementInvalid ...
+      return e if e.is_a?(NativeException) # JRuby 1.6
+      return e if e.is_a?(Java::JavaLang::Throwable)
+
+      case e
+        when ActiveModel::RangeError, SystemExit, SignalException, NoMemoryError then e
```
kares: is this seriously what AR is doing - they refactored but still ended up with mr. :hankey: ?

src/java/arjdbc/jdbc/RubyJdbcConnection.java
```diff
@@ -2016,132 +2019,84 @@ protected void setStatementParameters(final ThreadContext context,
...
-                final List<?> _param = (List<?>) param;
-                column = (IRubyObject) _param.get(0); param = _param.get(1);
-            }
-            else if ( param instanceof Object[] ) {
-                final Object[] _param = (Object[]) param;
-                column = (IRubyObject) _param[0]; param = _param[1];
-            }
-
-            setStatementParameter(context, runtime, connection, statement, i + 1, param, column);
+            setStatementParameter(context, connection, statement, i + 1, (IRubyObject) binds.get(i));
```
kares: if binds is an `RubyArray` it will attempt auto toJava conversion - which we might not want
... maybe we can just assume binds is always a RubyArray, right?

lib/arjdbc/abstract/database_statements.rb
rdubya: Unfortunately, they have a test that explicitly tests that it can still handle legacy bind parameters, so I'm not sure why its there but seems to be intentional for now.

```
    if ActiveRecord::Base.connection.prepared_statements
      def test_select_all_with_legacy_binds
        post = Post.create!(title: "foo", body: "bar")
        expected = @connection.select_all("SELECT * FROM posts WHERE id = #{post.id}")
        result = @connection.select_all("SELECT * FROM posts WHERE id = #{Arel::Nodes::BindParam.new.to_sql}", nil, [[nil, post.id]])
        assert_equal expected.to_hash, result.to_hash
      end
    end
```

lib/arjdbc/common_jdbc_methods.rb
rdubya: Most of this is copied over from the existing jdbc adapter. I'm not sure how much of it is still necessary, but I figured it was best to make the sqlite adapter have the same base translations as the other adapters.

I believe I had to add the `ActiveModel::RangeError` check because we don't typecast the values until we are in the Java code, at which point any error gets passed through here. The `RangeError` was then being wrapped as a sql exception.

src/java/arjdbc/jdbc/RubyJdbcConnection.java
rdubya: Yeah, as far as I can tell `binds` should always be a `RubyArray`. If we assume that, what changes should I make here?

lib/arjdbc/common_jdbc_methods.rb
kares: understood, thanks for the explanation - thought you copied over from AR, if not its fine as is

src/java/arjdbc/jdbc/RubyJdbcConnection.java
kares: would start with the RubyArray path that got deleted - just cast directly, for now and let the API cleanup happen later
~ `((RubyArray) binds).eltIternal(i)` ... so its clear what were passing here, guess we never passed anything else

rdubya:
@kares I updated it to assume that it is a RubyArray. Looks like everything is working the same as before so I think we're ok making that assumption.

kares:
@rdubya great, feel free to merge at your own as you prefer ...

PR #818, opened 2017-10-04
Make Postgres adapter a subclass of AbstractAdapter

rdubya:
This completes the work needed to make the postgres adapter a subclass of `AbstractAdapter` directly instead of extending `JdbcAdapter`, which made it so I could remove some of the clunky code that was needed to use the correct super methods.

I stopped including `ActiveRecord::ConnectionAdapters::PostgreSQL::DatabaseStatements` because most of the methods were being overridden anyways. I copied `exec_insert`, `explain` and `sql_for_insert` directly into the adapter since they were the only ones being used.

This also cleans up a few more failing tests.


lib/arjdbc/abstract/connection_management.rb
```diff
@@ -0,0 +1,22 @@
...
+      end
+
+      def disconnect!
+        super
+        @connection.disconnect!
+      end
+
+      def reconnect!
+        super
+        @connection.reconnect! # handles adapter.configure_connection
```
enebo: Doesn't this return @connection?

lib/arjdbc/common_jdbc_methods.rb
```diff
@@ -1,27 +0,0 @@
...
-
-      connection ||= jdbc_connection_class(config[:adapter_spec]).new(config, self)
-
-      super(connection, logger, config)
-    end
-
-    def translate_exception(e, message)
-      # we shall not translate native "Java" exceptions as they might
-      # swallow an ArJdbc / driver bug into a AR::StatementInvalid ...
-      return e if e.is_a?(NativeException) # JRuby 1.6
```
enebo: I don't think this NativeException line is needed if it is we should change the comment since JRuby 1.6 will never be able to run Rails 5.x

rdubya: I just copied this over from the jdbc adapter so I'm good with removing it if it isn't needed, I wasn't sure what it was catching so I didn't want to touch it :)

lib/arjdbc/abstract/connection_management.rb
rdubya: I don't know, this was another piece I copied over from the jdbc adapter, I'll see if that is what it is doing and update it if it does return it

lib/arjdbc/common_jdbc_methods.rb
kares: ... can pbly go for 5.x (needing JRuby 9K anyway) but the constant is still there so it won't fail :)

lib/arjdbc/abstract/connection_management.rb
rdubya: All the references to this that I could find didn't seem to expect a response, so I made it so this no longer returns the connection

rdubya:
I pulled out the couple of lines that were mentioned. Are there any other changes or is this good to merge?

rdubya:
@kares It looks like I introduced a syntax error in my previous PR that is fixed in this one. Is it ok for me to merge this so that the rails-5 branch is good again?

lib/arjdbc/common_jdbc_methods.rb
enebo: @kares yeah true I am just trying to remove potentially unknown-reason code from this branch.  If it is due to our pre-NativeException case in older JRuby we should remove it and just see what breaks.

enebo:
@rdubya I say land this because I am working on arjdbc today :)  Plus come on irc so I can ask you about getlocal Time in where clauses (sqlite3 breaks but I bet postgresql does not)

PR #820, opened 2017-10-13
Support Rails 5.0 prepared statement cache in the postgres adapter

rdubya:
This adds an abstract implementation for the prepared statement cache that was added in Rails 5.0.

I had to modify the connection setup so that the statements would actually be prepared on the server while running Rails tests. Let me know if there is a better way to handle that.


kares:
not sure I see a point of these with JDBC drivers ... personally, wouldn't care to implement this part. 
just more Ruby code to execute, while the drivers can usually do a much better job of PS caching, or?
but if you're for AR parity than sure let's have it.

lib/arjdbc/postgresql/connection_methods.rb
```diff
@@ -42,6 +42,10 @@ def postgresql_connection(config)
     properties['tcpKeepAlive'] ||= config[:keepalives] if config.key?(:keepalives)
     properties['kerberosServerName'] ||= config[:krbsrvname] if config[:krbsrvname]
 
+    # Need to force the prepareThreshold to 1 for AR unit tests otherwise
+    # we won't receive an error for the tests that check on the statement cache being invalidated
+    properties['prepareThreshold'] = 1 if database.start_with?('activerecord_unittest')
```
kares: let's not have code like these around, please :wink:

enebo: @kares This property needs to be set somewhere for running AR tests otherwise there is at least one failure.  Can you think of a good way of including this as a hook only when running AR tests?

rdubya: This change fixes 2 AR tests and unfortunately breaks another, but I think that is going to be one of the tests we just have to accept as broken at this point. I'd rather put it somewhere else but I'm not familiar enough with the test setup to know where else it can be inserted.

rdubya:
@kares from what I can tell, postgres doesn't actually cache statements unless it is told to. The jdbc driver defaults to only caching the statement on the server if the same PreparedStatement instance is executed 5 times. Currently, since the statement is closed at the end of each query, the statements never get cached on the server. It would be nice to move the cache into java but maybe we can bite that off once we try to get a version compatible with Rails 5.1.

Here are some references that I found about the caching:
http://use-the-index-luke.com/sql/explain-plan/postgres/concrete-planning
http://www.postgresql-archive.org/Monitoring-query-plan-cache-td5831561.html

I actually had a hard time finding complex queries that will trigger the cache but I made a small benchmark to test the difference and I think it is a big enough difference to keep the code in here. If a query isn't cacheable it shouldn't have the overhead of the additional calls back to Ruby. If you look at the "no cache enabled" the performance is basically the same as not using prepared statements at all:

## No prepared statements

```
Warming up --------------------------------------
         simple find    16.000  i/100ms
      multiple finds     5.000  i/100ms
Calculating -------------------------------------
         simple find    226.714  (± 6.2%) i/s -      1.136k in   5.033147s
      multiple finds     67.132  (± 6.0%) i/s -    335.000  in   5.011176s

Warming up --------------------------------------
         simple find    15.000  i/100ms
      multiple finds     5.000  i/100ms
Calculating -------------------------------------
         simple find    222.530  (± 7.2%) i/s -      1.110k in   5.016631s
      multiple finds     68.753  (± 4.4%) i/s -    345.000  in   5.031638s
```

## Prepared statements w/o cache

```
Warming up --------------------------------------
         simple find    16.000  i/100ms
      multiple finds     5.000  i/100ms
Calculating -------------------------------------
         simple find    223.951  (± 6.3%) i/s -      1.120k in   5.022622s
      multiple finds     68.680  (± 5.8%) i/s -    345.000  in   5.040767s

Warming up --------------------------------------
         simple find    15.000  i/100ms
      multiple finds     5.000  i/100ms
Calculating -------------------------------------
         simple find    224.171  (± 7.6%) i/s -      1.125k in   5.053163s
      multiple finds     68.697  (± 5.8%) i/s -    345.000  in   5.042328s
```

## Prepared statements w/ cache

```
Warming up --------------------------------------
         simple find    22.000  i/100ms
      multiple finds     8.000  i/100ms
Calculating -------------------------------------
         simple find    405.201  (±12.6%) i/s -      2.002k in   5.029363s
      multiple finds    102.619  (± 7.8%) i/s -    512.000  in   5.020565s


Warming up --------------------------------------
         simple find    21.000  i/100ms
      multiple finds     7.000  i/100ms
Calculating -------------------------------------
         simple find    376.770  (±10.6%) i/s -      1.869k in   5.019888s
      multiple finds    101.425  (± 6.9%) i/s -    511.000  in   5.060917s
```

Here is the benchmark code I used: https://gist.github.com/rdubya/36d63c61633826c0eb0461fcf3b51057 . I had to run it separate times for the different configurations.

lib/arjdbc/postgresql/connection_methods.rb
kares: are we seriously talking about having `database.start_with?` check out for `adapter: postgresql` :)
you probably have some context I am missing but let's not surprise users with hack such as those.

if its much needed to running Rails tests why not put it into the test adapter/database [configuration](https://github.com/jruby/activerecord-jdbc-adapter/blob/master/test/db/postgres_config.rb) ?

lib/arjdbc/abstract/database_statements.rb
```diff
@@ -12,7 +12,8 @@ def exec_query(sql, name = nil, binds = [], prepare: false)
           execute(sql, name)
         else
           binds = convert_legacy_binds_to_attributes(binds) if binds.first.is_a?(Array)
-          log(sql, name, binds) { @connection.execute_prepared(sql, binds) }
+          # It seems that #supports_statement_cache? is defined but isn't checked before passing "prepare" into here
+          log(sql, name, binds) { @connection.execute_prepared(sql, binds, prepare && supports_statement_cache?) }
```
kares: seems very fishy - if we have to `supports_statement_cache?` check here. 
> we didn't need it so far ... but I can imagine AR being opinionated :disappointed:

lib/arjdbc/postgresql/adapter.rb
```diff
@@ -816,6 +818,21 @@ def create_table_definition(*args) # :nodoc:
...
+      super
+    rescue ActiveRecord::StatementInvalid => e
+      raise unless e.cause.message.include?('cached plan must not change result type'.freeze)
+
+      if open_transactions > 0
+        # In a transaction, have to fail it - See AR code for details
+        raise ActiveRecord::PreparedStatementCacheExpired.new(e.cause.message)
+      else
+        # Not in a transaction, clear the prepared statement and try again
+        delete_cached_statement(sql)
```
kares: very ugly indeed ...

kares:
@rdubya thanks for the lengthy explanation - however I am still not 100% convinced its a good idea.

Rails does it because the drivers are really slim and each does its own piece, but since we have a common JDBC API (and drivers support configuring statement cache etc.) I somehow actively tried to avoid 'porting' such pieces over. what if you'd like to control the underlying layer e.g. with JNDI - this will stand in your way.

why is this needed in the first place - it's very "hacky" and _ugly_ on both .rb and .java ends ?
... aren't we simply pushing the Rails test mantra a bit too far due a single/few test failures ?

lib/arjdbc/postgresql/connection_methods.rb
enebo: @kares this is really only meant for a setup ONLY for running AR tests.  I think it was put in here initially due to lack of better location.

enebo:
@kares so my question is why does this nearly double performance?  Was there something wrong with rails-5 branch?  Or do users have to use something extra + JNDI to get this performance?  I do totally agree with you that if someone wants to use JNDI this should not interfere but it seems to give a better out of the box experience for default option users.  If there is a way we can get rid of ugliness while also allowing JNDI to work I think we should consider that.

My desire is to minimize the delta between Rails code and ours with the eventual goal of pushing a lot of Ruby back to being in AR itself.  I do not want to hobble advanced/nice features of JDBC though.  We are nearly 2 years into Rails 5 with only largely one person who was doing any support on it.  I feel like this move to be more Railsish will help with this issue too.

rdubya:
@kares @enebo Here's take 2. I pulled the statement cache management into the Ruby side of things. I made it so that now you can specify `statement_limit: 0` and it will disable the cache even if prepared statements are enabled. Also, I updated it so that it now passes rails tests an arjdbc specific config file so that we can better control connection options.

lib/arjdbc/abstract/statement_cache.rb
```diff
@@ -0,0 +1,58 @@
...
+      def clear_cache!
+        @statements.clear
+      end
+
+      def delete_cached_statement(sql)
+        @statements.delete(cached_statement_key(sql))
+      end
+
+      def fetch_cached_statement(sql)
+        @statements[cached_statement_key(sql)] ||= @connection.connection.prepare_statement(sql)
```
rdubya: I had originally created a method in RubyJdbcConnection.java to handle the statement preparation, but realized it wasn't really adding anything and meant more code to maintain. I can go back to using that if you think we should avoid directly referencing the connection object here.

Here's the method I had originally added:

```
    @JRubyMethod(name = "prepare_statement", required = 1)
    public IRubyObject prepareStatement(final ThreadContext context, final IRubyObject sql) throws SQLException {
        return withConnection(context, new Callable<IRubyObject>() {
            public IRubyObject call(final Connection connection) throws SQLException {
                final String query = sql.convertToString().getUnicodeValue();
                final PreparedStatement statement = connection.prepareStatement(query);
                return JavaEmbedUtils.javaToRuby(context.getRuntime(), statement);
            }
        });
    }
```

enebo:
I say land this and if we need to change things later we can.

PR #821, opened 2017-10-17
Fix multi-statement support

rdubya:
The fixes the failing test for multi-statement support and a couple of warnings about ambiguous arguments.

### DIFF
--- a/src/java/arjdbc/jdbc/RubyJdbcConnection.java
+++ b/src/java/arjdbc/jdbc/RubyJdbcConnection.java
@@ -555,14 +555,17 @@ public class RubyJdbcConnection extends RubyObject {
                     ResultSet resultSet = null;
 
                     while (hasResultSet || updateCount != -1) {
-                        resultSet = statement.getResultSet();
 
-                        if (resultSet != null) {
+                        if (hasResultSet) {
+                            resultSet = statement.getResultSet();
+
                             // Unfortunately the result set gets closed when getMoreResults()
                             // is called, so we have to process the result sets as we get them
                             // this shouldn't be an issue in most cases since we're only getting 1 result set anyways
                             columns = extractColumns(context.runtime, connection, resultSet, false);
                             result = mapToResult(context, context.runtime, connection, resultSet, columns);
+                        } else {
+                            resultSet = null;
                         }
 
                         // Check to see if there is another result set
--- a/src/java/arjdbc/jdbc/RubyJdbcConnection.java
+++ b/src/java/arjdbc/jdbc/RubyJdbcConnection.java
@@ -544,15 +544,49 @@ public class RubyJdbcConnection extends RubyObject {
 
                 try {
                     statement = createStatement(context, connection);
+                    boolean hasResult = doExecute(statement, query);
+                    int updateCount = statement.getUpdateCount();
 
-                    if (doExecute(statement, query)) {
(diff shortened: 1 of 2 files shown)


enebo:
@rdubya @kares I don't think AR supports this and unless we know people need this I think we should not support this.  If it nice that JDBC can support this but it potentially makes our support more complicated for a feature I suspect most Rubyist don't know exists...

enebo:
Also it looks like I broke this feature on rails-5 so perhaps reverting the commit which broke it will restore it.  I guess my question still stands from previous comment though.

kares:
agree with @enebo ... although I don't mind if you put this in - just make sure its not used by default.
hopefully in a separate method that does not get used much during default execution plan :) ... could be an optional feature that is turned on using a `-Darjdbc....` system property or simply a configuration option is set (not sure which one would make more sense). 

there definitely was demand on "beyond" AR features from users - e.g. unified `exec_proc` support which could have been done with JDBC having standard APIs for stored procedures.

kares:
> The fixes the failing test for multi-statement support and a couple of warnings

assuming this is for Postgres, in which case its probably safe (assuming the driver likes it) to have it but obviously not in the shared base class ... or did we misunderstand somewhere, does C PG handle this?

rdubya:
@kares @enebo Just to be clear ARJDBC 1.3 does support this, it was broken in the rails-5 branch. The tests that it fixes are arjdbc tests. ActiveRecord doesn't support this. It actually has inconsistencies in what happens if you do try to do this.
- AR's sqlite adapter -> returns the results from the first query in the string (presumably because sqlite only processes that part of the query, it appears to ignore anything after the semicolon)
- AR's postgres adapter -> returns a PG::Result with the results from the last query in the string
- AR's mysql adapter -> throws a StatementInvalid exception saying there is a syntax error near the second query in the string

Another inconsistency related to this is returning generated ids from an insert statement and results of an update statement. Currently ARJDBC supports returning the ids if the db supports it but AR adapters have different functionality:

Adapter | Insert | Update
-------- | ----- | --------
sqlite | empty array | empty array
postgres | PG::Result with number of rows inserted as cmd_tuples | PG::Result with number of rows updated as cmd_tuples
mysql | nil | nil


So I think we should drop support for multiple statements unless there is demand for it, since it only appears to be supported by MSSQL and Postgres (and maybe Oracle?) We should return the last result set found instead of the first one and deal with the results of an insert/update on a per adapter basis. It seems though that if a user is using `execute` instead of the action specific calls (`exec_update`, `exec_insert`, etc.) they can't really depend on being able to switch between DBs or ruby versions without having to make changes to the code.

If this makes sense I will update this PR to do that.

kares:
slight note: `execute` was never meant to be consistent but certain users (and we internally) might have dependend on its different behaviour (return) based on driver used. am not saying that it needs to be kept.

rdubya:
@kares @enebo Ok, at this point I have it so that we will only ever return the empty array or a result. I removed the test that was broken and I've updated it to return the last result from the set if a multi-statement query is made so that it is consistent with AR. This means that queries like `UPDATE pages SET title = 'New Page' WHERE id = 1; SELECT * FROM pages` will give the same result for the ARJDBC postgres adapter and the AR postgres adapter now.

I think we can worry about adjusting the return values if users run into issues with them, but there aren't any tests broken because of the return values as far as I can tell.

Issue #826, opened 2017-11-06
Binary string support when using prepared statements (AR 5.0 compat)


rdubya:
Adding this to the list with https://github.com/jruby/activerecord-jdbc-adapter/issues/824

There are currently tests marked as pending that test this functionality. We should get those working. Currently all bit string values are assumed to be booleans when they are set on the prepared statement.

PR #827, opened 2017-11-08
Fixed bit string support in postgres when using prepared statements

rdubya:
Previously bit types were being considered as booleans and would throw an exception when prepared statements were enabled.

Fixes https://github.com/jruby/activerecord-jdbc-adapter/issues/826


varyonic:
Did `jruby:rails-5` get deleted? Seems to be missing?

rdubya:
the `rails-5` branch is now the `master` branch

PR #829, opened 2017-11-12
(official) support for PG JDBC 42

kares:
42.0.0 was released after 9.4.1212 and does not contain any major breakage for AR-JDBC

current master (targeting AR 5.0) is fixed to not use 42 https://github.com/jruby/activerecord-jdbc-adapter/commit/7f4e5c1d8922d040ec6b65e6f51352c4f80599f4

... after this is tested we shall only support 9.4 / 42 versions (maybe even specify a more concrete 9.4)


enebo:
@kares If @rdubya lands this and it passes in his env as well as yours we can put this into 50.0 on monday morning.  If not then I will tag 50.0 to be right before this lands.

@kares a second observation is for rails 5+ it depends on JRuby 9.1 which is not Java 7+ so perhaps not all of those java specifications are needed?

kares:
its needed since AR-JDBC is not the only piece of SW using jdbc-postgres (e.g. Sequel uses the driver as well)
+ it might work on non 50 versions of AR-JDBC as/when needed

enebo:
@kares ok I did not know about sequel.  At some point, we may end up having to retire this since JRuby 1.7.x has been EOL for a while.  I do think no doubt some people are using it so I don't want to rain on anyone's parade :)

rdubya:
@kares @enebo I think the only issue I found was when trying to run the ARJDBC test suite, it fails with:

`NameError: uninitialized constant Jdbc::PostgresTest::SKIP_TESTS` in jdbc_postgres_test.rb

So, the suite won't run at all. I think we can probably merge this and just set that to false for now unless that should be defined somewhere.

The AR test suites seem to have pretty much all the same issues, and this driver update may have actually fixed a couple of failing tests (didn't try to go through and match them all up).

Thoughts on just merging it then patching that test file?

enebo:
@kares actually all our code gets built at the same time and is already using String switches.  I think we can correct this if need be but I will argue any gems using this new major version will just need to update to Java 7.  If there is enough pushback we can loosen back to 1.6 support (and if so decouple these two pieces of Java so one can be compiled 1.6 for backwards compay and the other can stay 1.7) but for now I say let's jettison this support.

PR #841, opened 2017-11-17
get rails tests to run (on CI)

kares:
previous tasks are left in ... `rake rails:test:sqlite3` still works

... new syntax is `rake rails:test_sqlite3` - uses AR-JDBC's dependencies, also faster to boot up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,6 @@ matrix:
         mariadb: '10.1'
       env: DB=mariadb PREPARED_STATEMENTS=true
     # Rails test-suite :
-    - env: DB=mysql2     TEST_PREFIX="rails:" AR_VERSION="5-0-stable"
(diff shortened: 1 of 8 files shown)


rdubya:
Looks ok to me, but I wonder if we should target the latest tag that we support instead of defaulting to the stable branch. The last time I tried that we had failing tests because of changes they had made in stable that hadn't been released yet, so that would make any PRs created in this project look like they broke the tests even though it was actually the tests that changed. Guess that could be a good thing though?

kares:
yeah, latest tag makes sense - will update. thanks for the feedback.

rdubya:
Was just taking a look at the ci build stuff. We should probably run the AR tests with all the same environments as the arjdbc tests. We'll at least want to be running them with prepared_statements on and off.

PR #843, opened 2017-11-21
Execute `configure_connection` to enable foreign keys at sqlite3 adapter

yahonda:
Refer https://github.com/rails/rails/pull/27701

This pull request addresses these two failures:

#### Steps to reproduce:
```
$ cd rails/activerecord
$ git checkout 5-1-stable
$ bundle install
$ ARCONN=jdbcsqlite3 bin/test test/cases/adapter_test.rb -n /test_foreign_key_violations/
```

### Actual result without this commit:
```ruby
$ ARCONN=jdbcsqlite3 bin/test test/cases/adapter_test.rb -n /test_foreign_key_violations/
Using jdbcsqlite3
Run options: -n /test_foreign_key_violations/ --seed 41207

# Running:

FF

Finished in 0.080637s, 24.8025 runs/s, 24.8025 assertions/s.

  1) Failure:
ActiveRecord::AdapterForeignKeyTest#test_foreign_key_violations_are_translated_to_specific_exception [/home/yahonda/git/rails/activerecord/test/cases/adapter_test.rb:288]:
ActiveRecord::InvalidForeignKey expected but nothing was raised.


  2) Failure:
ActiveRecord::AdapterForeignKeyTest#test_foreign_key_violations_are_translated_to_specific_exception_with_validate_false [/home/yahonda/git/rails/activerecord/test/cases/adapter_test.rb:278]:
ActiveRecord::InvalidForeignKey expected but nothing was raised.

2 runs, 2 assertions, 2 failures, 0 errors, 0 skips
$
```

Since the current `initialize` method does not call `configure_connection`  `disable_referential_integrity` method executes `PRAGMA foreign_keys = 0` at the end. which means foreign keys are disabled.

### DIFF
--- a/lib/arjdbc/sqlite3/adapter.rb
+++ b/lib/arjdbc/sqlite3/adapter.rb
@@ -78,6 +78,8 @@ module ArJdbc
 
       @active     = nil
       @statements = StatementPool.new(self.class.type_cast_config_to_integer(config[:statement_limit]))
+
+      configure_connection
     end
 
     def supports_ddl_transactions?


enebo:
@yahonda This is somewhat against my desires in that I want to keep that particular module as unchanged from AR as possible BUT I am already changing initialize because the arity does not fit in well with our codebase.  If you could change this to be in SQLite3Adapter at the bottom it would fit better but I will just add a comment that this is a difference.

yahonda:
Thanks for merging and I understand you wanted to keep module unchanged from ActiveRecord.

However, I do not fully understand yet what you mean "If you could change this to be in SQLite3Adapter at the bottom it would fit better but I will just add a comment that this is a difference."

Since the ActiveRecord `5-1-stable` and `master` branch, SQLite3Adapter initialize method calls `configure_connection` at the end. This structure is same as `ArJdbc::SQLite3` initialize method.

* `SQLite3Adapter`
https://github.com/rails/rails/blob/567f7037159d869bf0b903c4b6ccbca16d8926b7/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb#L95-L102

```ruby
      def initialize(connection, logger, connection_options, config)
        super(connection, logger, config)


        @active     = nil
        @statements = StatementPool.new(self.class.type_cast_config_to_integer(config[:statement_limit]))


        configure_connection
      end
```

Would you explain some more to help me to understand it. Thanks.

enebo:
@yahonda wow my mistake.  I must have missed this line.  I will remove that comment.  I had just diffed this the other day but I think I lost that line somehow.

yahonda:
No problem.  I understand your desires and agree with them.

PR #847, opened 2017-11-27
Let `supports_transaction_isolation?` have the same method signature

yahonda:
Let `supports_transaction_isolation?` have the same method signature with Active Record

This pull request the following failure.

```ruby
$ cd rails/activerecord
$ git checkout 5-1-stable
$ bundle install
$ ARCONN=jdbcmysql bin/test test/cases/transaction_isolation_test.rb:11
Using jdbcmysql
Run options: --seed 8415

F

Finished in 0.069170s, 14.4571 runs/s, 14.4571 assertions/s.

  1) Failure:
TransactionIsolationUnsupportedTest#test_setting_the_isolation_level_raises_an_error
[/home/yahonda/git/rails/activerecord/test/cases/transaction_isolation_test.rb:11]:
ActiveRecord::TransactionIsolationError expected but nothing was raised.

1 runs, 1 assertions, 1 failures, 0 errors, 0 skips
$
```

It also removes `test_supports_transaction_isolation` test expecting
`supports_transaction_isolation?` method have `level` argument, which is
not compatible with Active Record implementation.

Note: I am not clear how to handle Java implmentation for this method.

https://github.com/jruby/activerecord-jdbc-adapter/blob/b381039e78aed38ecb36e1ca8afbf137cc882865/src/java/arjdbc/jdbc/RubyJdbcConnection.java#L178-L198

### DIFF
--- a/lib/arjdbc/abstract/transaction_support.rb
+++ b/lib/arjdbc/abstract/transaction_support.rb
@@ -13,15 +13,8 @@ module ArJdbc
         @connection.supports_savepoints?
       end
 
-      # Does this adapter support setting the isolation level for a transaction?
-      # Unlike 'plain' `ActiveRecord` we allow checking for concrete transaction
-      # isolation level support by the database.
-      # @param level optional to check if we support a specific isolation level
-      # @since 1.3.0
-      # @extension added optional level parameter
-      def supports_transaction_isolation?(level = nil)
-        return false unless level
-        @connection.supports_transaction_isolation?(level)
+      def supports_transaction_isolation?
+        @connection.supports_transaction_isolation?
       end
 
       ########################## Transaction Interface ##########################
--- a/lib/arjdbc/mysql/adapter.rb
+++ b/lib/arjdbc/mysql/adapter.rb
@@ -54,6 +54,10 @@ module ActiveRecord
         true
       end
 
+      def supports_transaction_isolation?
+        true
+      end
+
       # HELPER METHODS ===========================================
 
       # Reloading the type map in abstract/statement_cache.rb blows up postgres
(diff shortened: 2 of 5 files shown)


kares:
this is (was) an AR-JDBC 'extension' - actually used it to check support for a given tx isolation level.
likely it was very specific and I might be the only 1 to `support_transaction_isolation?(:serializable)`

enebo:
@kares if we need this as a query can we make it something which will not conflict namewise with AR?  Possibly just make it for the one-level if so.

enebo:
@yahonda this should also get merged back to 50.x right?

yahonda:
>  this should also get merged back to 50.x right?

Yes. Here is the `supports_transaction_isolation?` method definition in Rails `5-0-stable` branch.


https://github.com/rails/rails/blob/dfb49730ab289d6bf3372bbaf78ed837cf6a0505/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L284-L287

```ruby
      # Does this adapter support setting the isolation level for a transaction?
      def supports_transaction_isolation?
        false
      end
```

kares:
actually I recall know - it was due embed adapters (e.g. Derby) ... fine grained checking was needed. 
anyways its possible to backport these to JdbcAdapter only as long as native piece stays intact.

enebo:
@kares if we do perhaps we can mix in JRuby-specific method names (which won't match AR ones) to isolate this behavior from the main adapters.  It is one thing I hope in the future is any extensions are obvious.  Maybe mixing in something like "jdbc/extensions/transaction_isolation" would help document it.  Just brainstorming in a random PR! :)

Issue #845, opened 2017-11-22
SSL configuration necessary to perform rake test_mysql


yahonda:
### Steps to reproduce:

I have created a rails-dev-box with JRuby support to test ActiveRecord JDBC Adapter. I can reproduce this error as follows.

* Vagrant Vagrant 2.0.1
* VirtualBox Version 5.2.0 r118431 (Qt5.6.2)

### At Vagrant Host

> git clone -b runs_ar_jdbc https://github.com/yahonda/rails-dev-box.git
> cd rails-dev-box
> vagrant up
> vagrant ssh

### At Vagrant guest

```
vagrant@rails-dev-box:~$ git clone https://github.com/jruby/activerecord-jdbc-adapter.git
vagrant@rails-dev-box:~$ cd activerecord-jdbc-adapter/
vagrant@rails-dev-box:~/activerecord-jdbc-adapter$ bundle
vagrant@rails-dev-box:~/activerecord-jdbc-adapter$ rake test_mysql
```

### Actual result
```ruby
"ActiveRecord::JDBCError: Trying to connect with ssl, but ssl not enabled in the server"
I did not configure SSL though, it looks this configuration recognizes JDBC driver as MariaDB one (:driver=>"org.mariadb.jdbc.Driver"), not MySQL.
```

Here are entire output of this command

```ruby
vagrant@rails-dev-box:~/activerecord-jdbc-adapter$ rake test_mysql
Using ActiveRecord::VERSION = 5.1.4
ActiveRecord::JDBCError: Trying to connect with ssl, but ssl not enabled in the server
                                                                                                           init_connection at arjdbc/jdbc/RubyJdbcConnection.java:424
                                                                                                                initialize at /home/vagrant/activerecord-jdbc-adapter/lib/arjdbc/jdbc/connection.rb:15
                                                                                                                initialize at /home/vagrant/activerecord-jdbc-adapter/lib/arjdbc/abstract/core.rb:13
                                                                                                                initialize at /home/vagrant/activerecord-jdbc-adapter/lib/arjdbc/abstract/statement_cache.rb:21
                                                                                                                initialize at /home/vagrant/activerecord-jdbc-adapter/lib/arjdbc/mysql/adapter.rb:34
                                                                                                           jdbc_connection at /home/vagrant/activerecord-jdbc-adapter/lib/arjdbc/jdbc/connection_methods.rb:15
[...]

PR #849, opened 2017-11-27
Address `sun.security.validator.ValidatorException:`

yahonda:
This pull request addresses #845 

```ruby
$ rake test_mysql
Using ActiveRecord::VERSION = 5.1.4
ActiveRecord::JDBCError: java.sql.SQLNonTransientConnectionException:
Could not connect to localhost:3306:
sun.security.validator.ValidatorException: PKIX path building failed:
sun.security.provider.certpath.SunCertPathBuilderException: unable to
find valid certification path to requested target
```

Refer [Using TLS/SSL with MariaDB Connector/J](https://mariadb.com/kb/en/library/using-tlsssl-with-mariadb-connectorj/)

### DIFF
--- a/lib/arjdbc/mysql/connection_methods.rb
+++ b/lib/arjdbc/mysql/connection_methods.rb
@@ -68,6 +68,7 @@ ArJdbc::ConnectionMethods.module_eval do
       # According to MySQL 5.5.45+, 5.6.26+ and 5.7.6+ requirements SSL connection
       # must be established by default if explicit option isn't set :
       properties['useSSL'] ||= false
+      properties['trustServerCertificate'] ||= true
     end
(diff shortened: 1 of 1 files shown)


kares:
actually, it is more of an issue of MariaDB not picking up `useSSL: false` (their alias stopped working)
... and also **rake test_mysql** using MariaDB's driver since the LOAD_PATH is messed, both of these are expected to be already fixed on **support-50-support** (which might take a week or so to lend on master)

... the problem with your fix is that you will replace an explicit `useServerCertificate: false` with `true`

yahonda:
Thanks for the update. I'm closing this and will wait for `support-50-support` merged to `master` 

Thank you.

Issue #852, opened 2017-11-30
rake test_sqlite3 gets "warning: loading in progress, circular require considered harmful"


yahonda:
Since  https://github.com/jruby/activerecord-jdbc-adapter/commit/da314bc1f7d54ef368e61658ea31953f0138e90f I am not able to run unit tests.

```ruby
$ cd activerecord-jdbc-adapter
$ bundle install
$ rake test_mysql
/home/yahonda/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/minitest-5.3.3/lib/minitest.rb:765: warning: loading in progress, circular require considered harmful - /home/yahonda/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/minitest-5.3.3/lib/minitest/test.rb
                                                                                                                   require at org/jruby/RubyKernel.java:955
                                                                                                                    <main> at /home/yahonda/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/minitest-5.3.3/lib/minitest.rb:765
                                                                                                                   require at org/jruby/RubyKernel.java:955
                                                                                                                    <main> at /home/yahonda/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/minitest-5.3.3/lib/minitest/test.rb:1
                                                                                                                   require at org/jruby/RubyKernel.java:955
                                                                                                                    <main> at /home/yahonda/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/minitest-5.3.3/lib/minitest/test.rb:1
                                                                                                                   require at org/jruby/RubyKernel.java:955
                                                                                                           block in (root) at /home/yahonda/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/bundler/gems/minitest-excludes-eefb471d2708/lib/minitest/excludes.rb:1
                                                                                                                      each at org/jruby/RubyArray.java:1734
                                                                                                                    (root) at /home/yahonda/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/bundler/gems/minitest-excludes-eefb471d2708/lib/minitest/excludes.rb:1
                                                                                                                    (root) at /home/yahonda/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/bundler-1.16.0/lib/bundler/runtime.rb:1
[...]

PR #855, opened 2017-12-01
Minitest excludes require nil

yahonda:
Follow up #854 and based on the comment https://github.com/jruby/activerecord-jdbc-adapter/issues/852#issuecomment-348529100


yahonda:
These changes have been made at master branch.

PR #857, opened 2017-12-15
[test] tune/more rails tests - leave PG config to its defaults

kares:
... seems that by doing so there's much more failures with Postgres' (rails tests) than currently

but I believe this is the test setup we should be going towards - explicitly not having a setup such as : 
```
production:
  insert_returning: ""
  prepared_statements: ""
```

maybe there's just one issue underneath the failures (check the rails:test_postgresql target on CI - the one without any ENV vars) otherwise it would mean the adapter is still not much in shape

### DIFF
--- a/test/rails/config.yml
+++ b/test/rails/config.yml
@@ -71,23 +71,25 @@ connections:
       prepared_statements: <%= ENV['PREPARED_STATEMENTS'] || 'false' %>
 
   postgresql:
-    arunit:
-      insert_returning: <%= ENV['INSERT_RETURNING'] %>
-      prepared_statements: <%= ENV['PREPARED_STATEMENTS'] %>
+    _setup: &_setup
+      <% if ENV.key?('INSERT_RETURNING') -%>
+      insert_returning: <%= ENV['INSERT_RETURNING'] %> # really should leave it off to no key if not set
+      <% end -%>
+      <% if ENV.key?('PREPARED_STATEMENTS') -%>
+      insert_returning: <%= ENV['PREPARED_STATEMENTS'] %> # should leave it off to its default if not set
+      <% end -%>
       username: <%= ENV['PGUSER'] || ENV['user'] || 'rails' %>
       properties:
+        # Postgres' JDBC driver does not prepare statements until executed 5 times by default, let's lower:
         prepareThreshold: <%= ENV['PREPARE_THRESHOLD'] || 1 %>
         # to not print (lengthy) traces from driver such as :
         #   SEVERE: Connection error: org.postgresql.util.PSQLException: FATAL: database "xxx" does not exist
         #loggerLevel: OFF # Broken in 42.1.4 thus let's do it manually :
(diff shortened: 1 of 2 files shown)


test/rails/config.yml
```diff
@@ -71,23 +71,24 @@ connections:
...
+      <%- if ENV.key?('INSERT_RETURNING') %>
+      insert_returning: <%= ENV['INSERT_RETURNING'] %> # really should leave it off to no key if not set
+      <%- end %>
+      <%- if ENV.key?('PREPARED_STATEMENTS') %>
+      insert_returning: <%= ENV['PREPARED_STATEMENTS'] %> # should leave it off to its default if not set
+      <%- end %>
       username: <%= ENV['PGUSER'] || ENV['user'] || 'rails' %>
       properties:
-        prepareThreshold: <%= ENV['PREPARE_THRESHOLD'] || 1 %>
+        prepareThreshold: <%= ENV['PREPARE_THRESHOLD'] %>
```
rdubya: We need the prepare threshold to be set to 1 for the tests that check some things with prepared statements otherwise the statements don't get prepared on the server like AR expects them to

enebo: @kares I do not have pgsql set up but I had to remove 'if' statements from a previous commit because I could not get rails to load the yaml: fb9e3a03d3fc83ed61914a95f49ed4a7d3e4bcc2.  I am not saying this is broken but I don't know why I had a problem then (when I am assuming you did not).

kares: @enebo ah I recalled it was in at some point - that is just a bit crazy ... we need to test w defaults

@rdubya so you think that a lot of failures are due this PS = 1? 
felt like the defaults are not handled right ... almost as if postgresql only passes 'reasonably' with the proper configuration in place (not a default configuration most user will have). 
its slightly scary to need to hack things to a 'non' real world setup. but o.k. I can leave it as is ...

rdubya: @kares yeah its a little odd but AR prepares the statement on the server the first time it is used and the postgres jdbc driver defaults to only preparing it on the server after the 5th time the prepared statement object is executed. In real world usage it wouldn't make much of a difference (other than not preparing the statements on the server for statements that are only used as 1 offs)

kares: aaah okey - thought its a limit for the statements prepared (upper bound for the driver pool).
... although the 5 vs 1 might still going to end up being interesting in real-world scenario :)

so I will put this back than, thanks

enebo: @kares @rdubya  hey let's add a comment there explaining it is JDBC not preparing until 5 so we don't have this conversation again in two years :)

kares:
updated `|| 1` with a comment, @enebo shall check if its working for him - so he won't undo this.
than we can figure out why there's more failures with the default config (compare the new CI targets).
there also should be a few fixes around PG adapter from https://github.com/jruby/activerecord-jdbc-adapter/pull/858

PR #858, opened 2017-12-15
backports (and cleanup) + redo returned time-values to match AR

kares:
this is rather big but most of the work are backports from *support* branch
(that branch used to be master but was force pushed over for the initial 50 release)

the support branch has its own issues (was unfinished territory) but also had several refactorings that would be quite a waste for AR-JDBC to leave behind + it was actually use by some folks :wilted_flower: 

most interesting (which I would personally also would like to see and will work towards getting in share) is being lazy to check-out JNDI connection and be able to `"configure_connection: false`
... which is slightly against how AR assumes to be working and required patches in 1.3 (JNDI support is currently still broken but when fixed there are already tests to check this is working - its mostly a task of properly undoing some of the work that went in to match up AR 5)

feel like the native Java side is good/clean enough at this point
planning on doing more workbut at this point it feels like it isn't breaking anything really ... and can do it in small incremental steps if this is on 50-stable and also merged over to master

one thing that was quite difficult and delayed this being 'ready' rather sooner is how we return time-like values -> for 50.0 this was 'changed' to return strings or resort to custom hacks (doing an 'optimistic' type lookup) in some cases. while the string option is decent, it breaks custom selects `SELECT now()` returning non type-casted result (might be considered a regression since it worked in 1.3 with AR 4.x).

those are now matched to work just like AR core -> where mysql2/pg are setup to handle Date/Time casting. so by the time those values reach AR's type system they don't adjust zone to match `AR.default_timezone` (like when they parse strings). tried to keep the (recent) commits around these with detailed info since its also tricky to test -> AR tests assume to be able to change system time-zone -> with the JVM and JDBC drivers isn't that easy.


lib/arjdbc/postgresql/oid_types.rb
```diff
@@ -105,30 +93,29 @@ def initialize_type_map(m)
...
         #   OID::SpecializedString.new(:interval, precision: precision)
         # end
         m.register_type 'interval', OID::SpecializedString.new(:interval)
 
-        m.register_type 'timestamp' do |_, _, sql_type|
-          precision = extract_precision(sql_type)
-          OID::DateTime.new(precision: precision)
-        end
+        register_class_with_precision m, 'time', Type::Time
+        register_class_with_precision m, 'timestamp', OID::DateTime
```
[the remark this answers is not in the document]
kares: think so - the changes here are mostly to match AR 5.0

lib/arjdbc/abstract/connection_management.rb
```diff
@@ -2,19 +2,33 @@ module ArJdbc
   module Abstract
     module ConnectionManagement
 
+      # @override
       def active?
+        return unless @connection
```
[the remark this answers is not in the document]
kares: it isn't 100% necessary - its useful to be able to distinguish a non-connected state
... its pbly a very rare state at this point (if ever happening)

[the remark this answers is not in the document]
rdubya: Yeah, seemed like something that shouldn't happen in regular use, but just so it has a consistent return type thought it would be good to return a boolean. Isn't there a different method to call to check if you are connected?

lib/arjdbc/abstract/core.rb
```diff
@@ -10,39 +10,26 @@ module Core
       def initialize(connection, logger = nil, config = {})
         @config = config
 
+        if self.class.equal? ActiveRecord::ConnectionAdapters::JdbcAdapter
```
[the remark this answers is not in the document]
kares: its quite ugly but with all the mixins the whole `super` setup is quite hard to get right.
\+ JNDI will complicate this a bit more - although that part should be mostly done here as well

since Core was extracted from JdbcAdapter (which used to be the base class 4 all) to be shared and due the order of supers I think this was the only way I got it to work ... the proper configuration for 'dynamic' adapter setups needs to be filled in early - its a legacy feature, but its not yet 'removed'.

it mostly seemed like the core extraction smh forgot about this as I reviewed the prev. changes.
we can remove it but I would do the ugliest thing - comment it with a note that if any adapter extending JdbcAdapter is brought to live for 5.x this needs to be un-commented ... otherwise it might take a few hours of figuring for the next person hitting this :smiling_imp:

lib/arjdbc/abstract/connection_management.rb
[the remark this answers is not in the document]
kares: its kind of expected to do both - believe AR's pool is using it.
starting to recall why it was done that way (this was changeg while cp-ing from `JdbcAdapter`) - that case was useful to distinguish whether to attempt a reconnect or not.
also Rails's default implementation returns `nil`. 
Rails isn't from the `strict?` true/false side is it? :wink:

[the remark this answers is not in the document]
rdubya: lol works for me then

lib/arjdbc/abstract/core.rb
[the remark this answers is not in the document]
rdubya: I'm fine with whatever you think is best, I'm just leery of code that checks to see if it is a specific class before doing something.

lib/arjdbc/jdbc/adapter.rb
```diff
@@ -528,24 +428,10 @@ def self.prepared_statements?(config)
...
-      # @private Supporting "string-subling" on AR 4.0 would require {#to_sql}
-      # to consume binds parameters otherwise it happens twice e.g. for a record
-      # insert it is called during {#insert} as well as on {#exec_insert} ...
-      # but that than leads to other issues with libraries that save the binds
-      # array and run a query again since it's the very same instance on 4.0 !
-      def suble_binds(sql, binds)
-        sql
-      end
+      # @deprecated no longer used
+      def suble_binds(sql, binds); sql end
```
enebo: We do not need to deprecate if this is dead or dying.  50.x is major release.

lib/arjdbc/oracle/adapter.rb
```diff
@@ -0,0 +1,863 @@
+# NOTE: file contains code adapted from **oracle-enhanced** adapter, license follows
```
enebo: @kares I think we decided to eliminate support for Oracle adapter since another gem supports Oracle for JRuby ruby-oci?

src/java/arjdbc/jdbc/JdbcConnectionFactory.java
```diff
@@ -1,45 +1,32 @@
-/***** BEGIN LICENSE BLOCK *****
- * Copyright (c) 2012-2013 Karol Bucek <self@kares.org>
- * Copyright (c) 2006-2010 Nick Sieger <nick@nicksieger.com>
- * Copyright (c) 2006-2007 Ola Bini <ola.bini@gmail.com>
+/*
```
enebo: Can this just go away?  Major release and is anyone outside really using this?

src/java/arjdbc/jdbc/RubyJdbcConnection.java
```diff
@@ -1677,7 +2161,7 @@ protected IRubyObject mapToResult(final ThreadContext context, final Ruby runtim
     }
 
     @Deprecated
-    protected IRubyObject jdbcToRuby(final Ruby runtime,
+    protected final IRubyObject jdbcToRuby(final Ruby runtime,
```
enebo: I am not a huge fan of final as a keyword in general but I really dislike final methods.  I would like the option of a subclass being able to override this method.

src/java/arjdbc/jdbc/RubyJdbcConnection.java
```diff
@@ -1756,43 +2240,57 @@ protected IRubyObject integerToRuby(final ThreadContext context,
...
+            if ( value == null /* || resultSet.wasNull() */ ) return context.nil;
+            return RubyString.newInternalFromJavaExternal(runtime, value);
+        }
+    }
+
+    protected static IRubyObject bytesToRubyString(final ThreadContext context,
+        final Ruby runtime, final ResultSet resultSet, final int column)
+        throws SQLException { // optimized String -> byte[]
+        final byte[] value = resultSet.getBytes(column);
+        if ( value == null /* || resultSet.wasNull() */ ) return context.nil;
```
enebo: I see this at least a couple of times.  Are you unsure about whether that || should be in there or not?  Either way I would say remove them and we can reinstall later if we need to.

src/java/arjdbc/jdbc/RubyJdbcConnection.java
```diff
@@ -2492,56 +2992,85 @@ protected void setObjectParameter(final ThreadContext context,
...
-    private IRubyObject setConnection(ThreadContext context, final Connection connection) {
-        close(getConnection(context, false)); // close previously open connection if there is one
-
-        final IRubyObject rubyConnectionObject = connection != null ?
-                convertJavaToRuby(connection) : context.runtime.getNil();
+    /**
+     * @note might return null if connection is lazy
+     * @return current JDBC connection
+     */
+    protected final Connection getConnectionImpl() {
```
enebo: another final method.

src/java/arjdbc/jdbc/RubyJdbcConnection.java
```diff
@@ -2693,51 +3267,46 @@ protected RubyArray mapColumnsResult(final ThreadContext context,
...
+            // (name, default, sql_type_metadata = nil, null = true, table_name = nil, default_function = nil, collation = nil, comment: nil)
+            final IRubyObject[] args = new IRubyObject[] {
+                columnName, defaultValue, type_metadata, nullable, tableName
+            };
+            columns.append( Column.newInstance(context, args, Block.NULL_BLOCK) );
         }
         return columns;
     }
 
+    @Deprecated
```
enebo: I might have missed others but I would like to just kill deprecated methods and start fresh.

src/java/arjdbc/jdbc/RubyJdbcConnection.java
kares: some drivers were fishy about the null checks - and since this is a base class for all ...
the original problem was that some of the wasNull check was rather expensive - unecessary
but than I am not sure at this point which/where ... the optimized string -> byte[] will be redone, I can remove this completely till than. was mostly a reference point to know it has been ported over.

src/java/arjdbc/jdbc/RubyJdbcConnection.java
kares: yep, thx - here it makes no sense really - will remove.
but "internal" (sensitive) pieces I would like to keep final

src/java/arjdbc/jdbc/JdbcConnectionFactory.java
kares: heritage of cherry-picking the backports from support can likely go - will do

lib/arjdbc/jdbc/adapter.rb
kares: heritage of cherry-picking the backports from support - will remove, thanks

src/java/arjdbc/jdbc/RubyJdbcConnection.java
kares: this one, I believe, should be fine - there's `getConnection` to override.
which than calls this piece - but this is a bit sensitive impl detail. 
makes no sense to override at this point.

kares:
> @kares At this point we have committed to sqlite3, mysql, postgresql, and mssql for 5{012...}.x. You added back Oracle support which I thought we agreed is already properly supported by another project. Unless we satisfy some need not supported by them I think we should kill Oracle support.

> You also added some code for Derby and Db2. If you think we can keep supporting those then I am ok with that since one person asked me at Rubyconf about DB2 (obviously someone wants it).

yeah - it should be stated that I did not try to work on these to support 5.x at all. 
those are simply due (cherry-picked) backports - likely due refactorings to keep the unit compilable.
if we want to kill it that should be addressed elsewhere, the motivation here was to have a solid (non-breaking) base of features that were done elsewhere and make sense for AR-JDBC's future.

regarding Oracle the enhanced adapter is very decent but could use a 'native' side for performance, which was a good opportunity here that wasn't delivered, not that I have any plans (fine with killing it off later).

btw. what I am seeing is Ruby-ist still being scared about Oracle as they attempt (J)Ruby on Rails/Oracle

thanks for the feedback, I will address your concerns.

src/java/arjdbc/jdbc/RubyJdbcConnection.java
enebo: ok

kares:
should look better now, let's ship this so we can push a few other fixes on top for next release 50/51 ...
unless there's a specific reason not to.

Issue #860, opened 2017-12-28
SQL injection possible in db2/as400 adapter


rsov:
Have an issue with sql injection in the db2 adapter. Ran the same code on SQL lite and DB2 and the results differ. Looks like SQL lite is getting rid of anything that's not a number when querying  a number field

```ruby
User.send(:sanitize_sql_hash, sin: "123123123/**/sql injection")
=> "\"users\".\"sin\" = 123123123" # SQL Lite
=> "\"users\".\"sin\" = 123123123/**/sql injection" # DB2
```

I am not sure how exactly rails sanitizes the sql but will look into the differences in source between sql lite and db2. I am also not sure how the sanitation works for the string fields.

Can be recreated by running this code: `User.find_by_sin('777888999/**/union...')`

Any help where to start looking will be appreciated

PR #861, opened 2017-12-28
Refactor db2 quote

rsov:
This is the solution to the issue #860

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,25 +46,6 @@ branches:
     - /^test-.*/
 matrix:
   allow_failures:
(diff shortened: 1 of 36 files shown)


kares:
Thanks Ruslan, sorry it tooks so long. Was hoping I will look in detail trying to understand the issue.
Giving up at this point simply due lack of my time to do so, trusting your solution by the gut :) ... thx!

rsov:
Thank you, Kares!

We have been running this code in prod for few months now on Rails 4.2. No issues so far

Issue #846, opened 2017-11-22
Geometric types in Postgres adapter aren't fully compatible 


rdubya:
There are a number of tests failing related to geometric types for the postgres adapter. Some fail with prepared statements both on and off, others are one or the other.

PR #863, opened 2018-02-05
Fix geometric type support in postgres adapter for AR 5.0

rdubya:
If this is good to go I'll also get it into master for AR 5.1 support. The main issue that I had to work around was that AR allows most of the formats that postgres allows when defining the geometric types but the PGobject classes were more strict on the strings they would accept. This fixes https://github.com/jruby/activerecord-jdbc-adapter/issues/846 and makes it so the geometric type tests are now passing.


rdubya:
Hey @enebo, have you gotten a chance to take a look at this?

enebo:
@rdubya looks reasonable to me.  Nothing stands out but hopefully this makes something pass tests.

rdubya:
@enebo thanks for taking a look. I think this fixes all the rails tests related to geometric types.

PR #866, opened 2018-02-20
Supply a non-nil column set

### LINKED ISSUE #865: exec_insert returns a result where most methods raise errors
The return result from `exec_insert` on an MySQL adapter is an `ActiveRecord::Result` constructed in an [unusual way](https://github.com/jruby/activerecord-jdbc-adapter/blob/4ae8a3c4d4b00d1c33ef6b9099f63a3f78da21ad/lib/arjdbc/mysql/adapter.rb#L109), such that it has `nil` for the set of columns. The result is that most methods on the returned `Result` instance (including `first`) raise an exception when invoked because most methods call the `hash_rows` method internally, which depends on `@columns` having a non-nil, non-empty value.

Observed locally with activerecord v5.0.6 and activerecord-jdbc-adapter 50.0, but the unusual result construction appears in the `master` branch as well.

Workaround is to call `rows.first` to retrieve the last insert id from the first row explicitly, rather than calling `first`.

JasonLunn:
Fixes #865

### DIFF
--- a/lib/arjdbc/mysql/adapter.rb
+++ b/lib/arjdbc/mysql/adapter.rb
@@ -106,7 +106,7 @@ module ActiveRecord
                   end
         # FIXME: execute_insert and executeUpdate mapping key results is very varied and I am wondering
         # if AR is now much more consistent.  I worked around by manually making a result here.
-        ::ActiveRecord::Result.new(nil, [[last_id]])
+        ::ActiveRecord::Result.new(['last_id'], [[last_id]])
       end
       alias insert_sql exec_insert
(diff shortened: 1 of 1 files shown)


enebo:
Excellent.  There is something we should be doing differently in Java space but the FIXME above that line is meant to indicate whatever we call through have many different possible return types.  So my original stab was mostly trying to make AR tests pass.  With that said, I am now wondering if we have other queries like this where we return nil for columns?

enebo:
haha whoops I thought I clicked merge vs close :)

PR #868, opened 2018-02-26
Pass column types for postgres results

rdubya:
This makes it so we will fetch the column types from the adapter to populate the `ActiveRecord::Result` object if we have access to them. It's a bit of a chicken/egg situation since we need to get the types from the cache but need to query to get types into the cache, so there may be a better work around than I currently have in place.

This fixes a couple of tests and sets us up to potentially fix a few others. I think it should also be applied to other adapters at some point. If this is ok, I'll see about getting it into master as well so it can be in the 51.0 release.

I created `PgResultSetMetaDataWrapper` because the `getField` method is protected in the `PgResultSetMetaData` object and it was the only way I could figure to work around it. If there is a better option please let me know.

### DIFF
--- a/lib/arjdbc/postgresql/oid_types.rb
+++ b/lib/arjdbc/postgresql/oid_types.rb
@@ -40,6 +40,11 @@ module ArJdbc
       end
 
       def get_oid_type(oid, fmod, column_name, sql_type = '') # :nodoc:
+        # This can be called before the type map is created
+        # For example, when querying the database version to find out what
+        # should be in the type map
+        return unless type_map
+
         if !type_map.key?(oid)
           load_additional_types(type_map, [oid])
         end
@@ -162,7 +167,12 @@ module ArJdbc
           query += initializer.query_conditions_for_initial_load(type_map)
         end
 
+        # Make it so the type map isn't available to prevent
+        # loops with get_oid_type
+        saved_type_map = @type_map
+        @type_map = nil
         records = execute(query, 'SCHEMA')
+        @type_map = saved_type_map
         initializer.run(records)
       end
 
--- a/src/java/arjdbc/jdbc/RubyJdbcConnection.java
+++ b/src/java/arjdbc/jdbc/RubyJdbcConnection.java
@@ -3644,7 +3644,7 @@ public class RubyJdbcConnection extends RubyObject {
         return Result.newInstance(context, columnsToArray(context, columns), rows, Block.NULL_BLOCK); // Result.new
     }
 
-    private static RubyArray columnsToArray(ThreadContext context, ColumnData[] columns) {
+    protected static RubyArray columnsToArray(ThreadContext context, ColumnData[] columns) {
         final IRubyObject[] cols = new IRubyObject[columns.length];
 
         for ( int i = 0; i < columns.length; i++ ) cols[i] = columns[i].getName(context);
@@ -3752,6 +3752,7 @@ public class RubyJdbcConnection extends RubyObject {
         public final int type;
 
         private final String label;
+        private IRubyObject rubyType = null;
 
         @Deprecated
         public ColumnData(RubyString name, int type, int idx) {
@@ -3768,6 +3769,11 @@ public class RubyJdbcConnection extends RubyObject {
             this.index = idx;
         }
 
+        public ColumnData(String label, int type, int idx, IRubyObject rubyType) {
+            this(label, type, idx);
+            this.rubyType = rubyType;
+        }
+
         // NOTE: meant temporary for others to update from accesing name
         ColumnData(ThreadContext context, String label, int type, int idx) {
             this(label, type, idx);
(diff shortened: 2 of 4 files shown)


lib/arjdbc/postgresql/oid_types.rb
```diff
@@ -40,6 +40,11 @@ def lookup_cast_type(sql_type)
       end
 
       def get_oid_type(oid, fmod, column_name, sql_type = '') # :nodoc:
+        # This can be called before the type map is created
```
kares: this is a good reason to have that one piece native (believe it was previously)
... and than hopefully no hacks shall be needed

kares:
honestly I am not sure about this being the right approach, here's how I see the progress so far for 5.x :
1. get rid of customizations (that might have been in native) and bring in adapter .rb from core
2. than we try small hacks in .rb and eventually push those down to native :)

would like to understand why core does not have this chicken-egg problem 
or why they haven't been needed in 1.x releases

... maybe, if its one call to get the version, we can work-around it by using the driver's API to get version?
what do you think, did I miss (or misunderstood) something? usually do :smile:

rdubya:
Hey @kares thanks for taking a look.

The reason that we haven't had the chicken/egg problem in the past is that we don't supply the types at all in the AR::Result object at this point so we don't ever try to look up what the type should be. Up until now we have been dependent on the type being reported by JDBC (i.e. `Types.Date`) and trying to do a conversion based on that so that it comes back as the right type and we don't need AR to know what type it was. This breaks down when trying to handle arrays though. Since we need to know what the internal type of the array is, detecting that it is `Types.ARRAY` doesn't give us enough information to properly deserialize the array items. An example of this is arrays of type json, we return an array of strings instead of an array of deserialized json objects. I tried to update `objectToRuby` to detect if it is of type json and deserialize it in that case, but then we get the wrong response for columns of type json (not arrays) because AR is expecting a string in that case instead of a deserialized object.

I tried testing for the version query to avoid the chicken/egg issue, but that just made it so I got past that one to the check for whether settings were enabled or not and failed there. So, we can try to catch all of those cases, but I feel like that would just be a place we'd be constantly finding new exceptions we need to add.

I think native AR does not have this issue because they return a `PGresult` object from `#execute` which doesn't use typing while we run `exec_query` and `execute` through the same path and return an `AR::Result` object in both cases. So, the best route is probably to separate those paths but I'm not  sure it makes sense to do in 50 or 51.

Maybe the best thing to do is just sit on this and try to get it into 52 (including changes to the query paths) and then backport it? At this point I was just trying to clean up a few things and then I'd like to focus on getting rails 5.2 support working. It would be nice if we could release 52.0 around RailsConf time.

lib/arjdbc/postgresql/oid_types.rb
rdubya: Do you mean "native" as in native AR or "native" as in Java based? This comment just made me realize that maybe it would be better to have this act more like AR in that the PG gem always returns a `PGresult` object and then its converted to an `AR::Result` object in `#exec_query` instead of having the java side create the `AR::Result` object directly

enebo: @rdubya In general we do not want to double box data if we can avoid it.  Directly generating proper AR::Result seems best so long as we have enough knowledge to do that.

src/java/arjdbc/jdbc/RubyJdbcConnection.java
```diff
@@ -3752,6 +3752,7 @@ protected static RubyString cachedString(final ThreadContext context, final Stri
         public final int type;
 
         private final String label;
+        private IRubyObject rubyType = null;
```
enebo: @rdubya I assume this is accessed from Ruby somewhere?  I don't see any consumers.

src/java/arjdbc/postgresql/PostgreSQLRubyJdbcConnection.java
```diff
@@ -280,6 +286,56 @@ protected Connection newConnection() throws SQLException {
...
+    protected static IRubyObject newResult(final ThreadContext context, ColumnData[] columns, IRubyObject rows) {
+        final RubyClass Result = getResult(context.runtime);
+        return Result.newInstance(context, columnsToArray(context, columns), rows, columnTypes(context, columns), Block.NULL_BLOCK); // Result.new
+    }
+
+    private static RubyHash columnTypes(ThreadContext context, ColumnData[] columns) {
+        RubyHash types = RubyHash.newHash(context.runtime);
+
+        for ( int i = 0; i < columns.length; i++ ) {
+            IRubyObject type = columns[i].getRubyType();
```
rdubya: @enebo This is where the `rubyType` you asked about is used

lib/arjdbc/postgresql/oid_types.rb
rdubya: I'll see what it turns into, this PR doesn't change how the result is generated so this should be ok.

src/java/arjdbc/postgresql/PostgreSQLRubyJdbcConnection.java
rdubya: This ultimately gets sent to the `AR::Result` object which is what uses the type object

rdubya:
I decided just to close this for now and do it all as batch so we don't have to put in the workarounds for the race conditions.

PR #869, opened 2018-03-06
fix schema config for Oracle application users

jkuchta:
Oracle application users are currently broken in 1.3.x, the SELECT that tries to determine the proper sequences and primary key uses a nil owner and fails. This is a quick fix that will set the schema_owner to the schema name if set.

### DIFF
--- a/lib/arjdbc/oracle/adapter.rb
+++ b/lib/arjdbc/oracle/adapter.rb
@@ -921,9 +921,13 @@ module ArJdbc
     # default schema owner
     def schema_owner(force = true)
       unless defined? @schema_owner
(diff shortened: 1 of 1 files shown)


kares:
any reason a piece of test was left out here?

jkuchta:
@kares I looked through the tests but didn't could find any place where adapter specific functionality was tested (Oracle). For such a small change on an older version, I don't have the time to create an entire Oracle testing suite.

kares:
its https://github.com/jruby/activerecord-jdbc-adapter/tree/1-3-stable/test/db/oracle

kares:
okay, fair enough. will take your word for it :wink:

PR #870, opened 2018-03-08
Align Java query interface with AR's

rdubya:
This goes back to calling Java methods that are more closely aligned with their associated AR methods (exec_query, exec_insert, etc)

This is basically a clean up to prep for making `#execute` start returning more of a raw result instead of `AR::Result` objects.


kares:
pretty much welcome these changes ... here's my 2 cents of feedback : 
- if this is only master - native changes will be harder to manage - merge/pick across branches
  ... you will pretty much "drop" (make it harder to support) 50.0 before a stable release
  (maybe this is wanted at this point, not sure, as I only managed to support multi from a single piece)
- if anyone wants to build on top of AR-JDBC this is making it harder by not having any Java API
  ... its easy to get these "right" for 'good' drivers such as those 3 but beyond there should still be a reasonable/stable API and this is pretty much only a bunch of native methods that work for those with fairly well behaving JDBC with less possibility of re-using if an ext goes native

rdubya:
Hey @kares, thanks for the thoughts. 

For your first point, I plan on merging this back to the stable 50 branch once it is in to master, I believe it should be a fairly straightforward merge.

On the second point, it seems like at this point we are fairly tightly integrated with active record, so I think anybody wanting to build other drivers would likely want to match AR's interface anyways. I'm currently working on making `execute` return a more standard result as well which they would also be able to use if these helper methods don't fit their model.

I don't really have a sense of what people would be doing with this outside of using it with active record so I'm certainly open to learning more on that front and trying to come up with a standard api that would work for them.

rdubya:
@kares, just realized you were probably referring more to people building additional db drivers than working with other ruby frameworks lol. I think for now its good to simplify things so we can reduce maintenance overhead and then if others do need the additional hooks we can add them back in.

enebo:
I also think using the Ruby base modules we have should become part of any new adapter work vs just leaning on the Java APIs.  If those do not work in conjunction with the Java side for new adapter support then we have an issue to work through at that point.  Those common modules are fairly central simplifying how we connect to AR on the Rails side.

Another rationalization is we still do not have enough contributors to be thriving with the adapters we have.  I don't want to limit possibilities but I also don't want to limit the support we can provide today by complicating what we are committed to atm.

kares:
agree but there are few ext out there - my point was solely not to make it "harder" for them.
if we kept some of the APIs it might have been easier (motivating) for them to eventually upgrade.
than again hopefully they're not depending on it much and I do not expect us to review those ...

enebo:
@kares is there a way we can find what these exts are out there?  Without concrete knowledge if what is used I don't think supporting all our variations because some ext might be using it is a problem.   Keeping all our APIs in Java we have today would be much more palatable if we knew who used them and why.  Maybe we can do a github search?  

I think the largest issue I have encountered with Rails 5 work is multiple permutations of similar methods without any personal knowledge of why those permutations are needed.  I know part of that answer was to support multiple versions of Rails, but I am also learning some of it involves Javay/JDBCy aspects.  So anything we can do to at least figure out what/why and then document it so in the future we are not wondering why a particular variant of a method exists.

PR #871, opened 2018-03-09
Increment index when counting octal escapes

elcubo:
This is to avoid arrayIndexOutOfBoundsException in the main unescape loop.  It only happens when unescaping non-escaped content.

### DIFF
--- a/src/java/arjdbc/postgresql/ByteaUtils.java
+++ b/src/java/arjdbc/postgresql/ByteaUtils.java
@@ -89,6 +89,7 @@ abstract class ByteaUtils {
                     }
                     else {
                         correctSize -= 3;
+  						i += 2;
                     }
                 }
             }


donv:
Tests run as well as before, and the spec of the method is the same as before, so I feel a test is unnecessary.  Merging.

donv:
Yes, changes with tests are better.  Do you know why this method is called with non-escaped content?

kares:
well that is what I was hoping I would get explained here ... there's no context established for the issue :disappointed:

donv:
Well, the larger context is that in a real-world app, this unescape method was being called by AR with a non-escaped value when loading a PG bytea value.  The fix was to avoid the exception and get the app working again.

After the concrete fix, I did some tracing, and I found that core AR+PG calls unescape on the column value without checking if it is actually escaped.  Since this seems to work, I expect there to be some mechanisms that handles this, but I was unable to find it.  So maybe this only works by coincidence.

At the time I guessed that the current state is due to ARJDBC moving towards removing custom code and use more AR core code, and that the core AR code depends on `pg` which I guess returns escaped content.  Many guesses here.

I'd be happy to look more into this, but I thought I'd ask if anyone knows why AR is unescaping the non-escaped value first.

kares:
@donv great, thanks. what would have avoided the confusion is either a test-case or at least some "vague" reproducer with a back-trace.
so that with changes in AR-JDBC we would have confirmed whether its not due some other issue really.
and we would also know whether it needs to get backported to 50.x or not. so I am leaving this as is for now.

rdubya:
@kares it looks like these changes may have disappeared because of that cherry picking that I did. Should I get them put back in to both branches? Based on your last comment I wasn't sure if they were supposed to be in there or not.

kares:
well I am not sure either and since there isn't any test/trace I would leave it as is. no tests fail and I am not sure why a non-escaped content would be attempted to unescape. this might hide a different bug ...

PR #873, opened 2018-03-21
Fix a TypeError ("can't convert nil into String") that happens when offset is used without limit

betesh:


betesh:
@kares I'd very much appreciate a new 1.3 release so I don't have to use github for this.  Thanks!

PR #874, opened 2018-03-22
auto (magically) build ext adapter_java.jar

kares:
this isn't something completely new - has been used on master before
I mean it was working fine with `:git` sources pointed towards AR-JDBC.

gist of it is that the adapter_java.jar is removed from repo and git ignored.
on release (there's a rake task for that) the adapter .jar is `git add -f` and commited.

the problem is that it seems to me that we're forgetting to commit the .jar
even if this is fixed eventually, as there are native changes, per branches it might cause headaches.

anything against this piece? really would prefer having to now worry about missed .jar commits


kares:
@enebo @rdubya any objections?

enebo:
@kares So rake will commit it but only when we release.  Otherwise we ignore it and reduce clutter of making our git repo slowly huge?  Sounds good if that is it.

An alternative may be to just require that anyeone who works on our project must just have javac installed and we just ignore it all the time?  That seems a lot simpler with the caveat that we inched the bar up a little bit for new developers.

rdubya:
I think it makes sense. I was never sure if I should be committing it on every change or not, and if we forget then it then it messes with the tests.

I could see skipping it altogether like @enebo mentioned too. Guessing most people trying to run this stuff would already have the jdk installed anyways, but it would potentially raise the bar a bit.

kares:
> An alternative may be to just require that anyeone who works on our project must just have javac installed and we just ignore it all the time?

well, its pretty much ignored all the time, but when releasing you want to include the piece so you do not trigger an ext build when installing the gem (but I see how it might be consufing). if I got it right than yes a full exclude is doable - include that one file despite it not being in the repo on release, guess we won't miss being able to lookup a tag. will look into that as a follow-up.

activerecord-jdbc-adapter.gemspec
```diff
@@ -9,7 +9,7 @@ Gem::Specification.new do |gem|
   gem.authors = ['Nick Sieger, Ola Bini, Karol Bucek and JRuby contributors']
   gem.email = ['nick@nicksieger.com', 'ola.bini@gmail.com', 'self@kares.org']
   gem.homepage = 'https://github.com/jruby/activerecord-jdbc-adapter'
-  gem.license = "BSD"
+  gem.license = 'BSD-2-Clause'
```
enebo: Will anyone perceive this as a license change or was BSD by itself just too ambiguous?

enebo:
@kares it will be double plus super saiyan awesome to get rid of that jar being committed!!!

activerecord-jdbc-adapter.gemspec
kares: RG was hitting me with a warning that I need to pick smt, as I tested building the gemspec

kares:
@enebo actually that is done, already. but thanks for the encouragement :)

PR #875, opened 2018-03-22
WIP: Provide a standard Result object

rdubya:
This creates a standard JdbcResult object that can be returned for "raw" results similar to the driver objects for the other adapters (i.e. PG::Result)

It moves all of the Java -> Ruby object conversion into dedicated column type classes. This helps to shrink down the size of the adapter classes as well as make it a little easier to follow where things are happening and what adapters are using what conversion types (I hope lol)

These changes fix a couple of rails tests and should be able to be back ported to the 50-stable branch. 

I plan on converting the mysql and sqlite adapters as well but wanted to get this up so that it could be reviewed before I spent waayyyy too much time on it. :)


src/java/arjdbc/postgresql/PostgreSQLResult.java
```diff
@@ -0,0 +1,173 @@
...
+ */
+public class PostgreSQLResult extends JdbcResult {
+
+    // These are needed when generating an AR::Result
+    private IRubyObject adapter;
+    private PgResultSetMetaData resultSetMetaData;
+
+    /********* JRuby compat methods ***********/
+
+    private static RubyClass rubyClass;
```
rdubya: I wasn't quite sure what all needed to happen to make these `IRubyObject`s and to be able to include the `Enumerable` module in it. So if there is a better way/less code way to handle it I'm all ears.

rdubya:
Another thing that came up while doing these changeovers: If I leave arrays and time/timestamps as strings, it actually fixes some postgres related test failures. Do we need to try to parse these? I'm not sure we actually have enough data to parse the arrays reliably enough, so I think we should leave those as strings. But, can we do the same thing for dates/times?

kares:
so this essentially means more native pieces to manage on our hand -> eventually less core .rb reuse.
and I somehow had the impression, previously, that we're not going towards that... quite the opposite.

could you maybe elaborate on what tests this resolved - just to be sure we're not fixing a few edge cases?

if it's motivated by `PG::Result` for Postgres' compatibility than OK (although would simply exclude those specific Rails tests - AR consumers never face PG native types so those tests might be internal) although other native drivers don't "leak" out their internals. sounds to me like we'll end up with 1 too many type systems - JDBC types, AR::Type and now AR-JDBC's internal types :) + PG adapter has its OID types.

TL/DR really understand PG is ugly as it is (lots of types) and it makes sense there but I feel like a whole new type-system for others might be an over-kill. on the other hand there's way to register type-handlers directly with the PG driver - would have the same code simplification effect.

... guess you guys won't feel like having a release without this (or only for PG) and introducing more later ?

we're also be doing slightly more allocation (or lookup) with these types on the path so I wonder how much impact extracting results for a table with many rows it has. this is more of a personal goal of mine - historically AR-JDBC has always been behind MRI performance, we got close at some point along 4.2 and are also hopefully pretty close by now (maybe except for string transcoding which is also solvable). 

had always viewed the native bits as a necessity that should rather be ugly but speedy than proper OO :smile:

rdubya:
Some of the tests this fixes are ones that are failing because ActiveRecord doesn't have the type information when the AR::Result is returned. It also had the side effect of fixing some of the bc timestamp issues because now it returns a string for dates, not sure if that is a good tradeoff or not. It also fixed some tests that were only failing because they were using PG::Result methods and it seems like trying to keep a somewhat consistent result between the two drivers would be good anyways for anybody trying to switch back and forth between MRI and JRuby.

I haven't done any performance testing, but it doesn't seem like we are allocating much more than we already were. Previously we created a `ColumnData` object for each column and now that is being replaced with `AbstractType` so we're creating the same number of objects for the column typing stuff. We then call a method on those objects instead of calling one in the adapter itself, so I guess that could add some overhead. Previously we were creating an `AR::Result` to return from `execute` which doesn't match the built in adapters which return the native result from the driver in that case. So this change helps the compatibility there.

Can you go into a little more detail about the performance concerns? I'm also a little confused about your comment about having the 3 different types, don't we essentially already have that? The arjdbc "types" are just defined as methods right now instead of objects, right?

I definitely don't have the Java experience that you and @enebo have so if this is a bad direction to go then let me know, but I think we should at least change to returning some sort of result object that isn't an AR::Result just so we can match up better with the expected functionality. From the AR code where `execute` is defined:

```
# Executes the SQL statement in the context of this connection and returns
# the raw result from the connection adapter.
# Note: depending on your database connector, the result returned by this
# method may be manually memory managed. Consider using the exec_query
# wrapper instead.
```

src/java/arjdbc/util/ActiveRecord.java
```diff
@@ -0,0 +1,36 @@
+package arjdbc.util;
+
+import org.jruby.RubyClass;
+import org.jruby.RubyModule;
+import org.jruby.runtime.ThreadContext;
+
+/**
+ * Provide a standard place to get references to ActiveRecord classes and modules
+ */
+public class ActiveRecord {
```
rdubya: Not sure on the naming of this one. I kind of like it because now you can call `ActiveRecord.Base(context)` and it mimics the ruby namespacing. I could see it being confusing though and it kind of goes against Java coding conventions.

rdubya:
This PR should actually reduce the number of objects that are created when processing a result since we no longer create a ColumnData object for each column.

src/java/arjdbc/util/ActiveRecord.java
enebo: I dislike this type largely because although they are static methods we continuously need to look these up in constant maps over and over.  Older code also did this but making this class I think will cement this design.  I would rather our RubyJdbc java class receive these as field in the constructor or somehow save them once they are retrieved (but they cannot be in static fields or we will create memory leaks of the whole ruby runtime0.

PR #876, opened 2018-03-26
Build adapter jar when installing from source

donv:
Since the  `adapter_java.jar` file is no longer committed to the repository (which is good), developers wanting to use ARJDBC from source need to build the jar before using the gem.

To make this easier we can add this build step as an optional java extension in the gemspec.

There was a check on RUBYARCHDIR in the Rake task for the adapter jar which used to be set by the build script.  This is not set anymore, so I removed the check.

I am not aware of any tests for this toolchain.  Feel free to suggest where I could add some.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -227,7 +227,7 @@ if defined? JRUBY_VERSION
       driver_jars
     end
 
-    if ENV['RUBYARCHDIR'] || ENV['BUILD_EXT_MAVEN'] == 'true'
+    if ENV['BUILD_EXT_MAVEN'] == 'true'
       driver_jars = get_driver_jars_maven.call
     else
       driver_jars = get_driver_jars_local.call
--- a/activerecord-jdbc-adapter.gemspec
+++ b/activerecord-jdbc-adapter.gemspec
@@ -37,5 +37,7 @@ Gem::Specification.new do |gem|
   #gem.add_development_dependency 'mocha', '~> 0.13.1'
 
   gem.rdoc_options = ["--main", "README.md", "-SHN", "-f", "darkfish"]
+
+  gem.extensions << 'Rakefile' if File.exist?('.git')
 end
 


activerecord-jdbc-adapter.gemspec
```diff
@@ -37,5 +37,7 @@ Gem::Specification.new do |gem|
   #gem.add_development_dependency 'mocha', '~> 0.13.1'
 
   gem.rdoc_options = ["--main", "README.md", "-SHN", "-f", "darkfish"]
+
+  gem.extensions << 'Rakefile' if File.exist?('.git')
```
kares: do not like where this is going - we could easily leave such piece in for the released gem, which would be bad

Rakefile
```diff
@@ -227,7 +227,7 @@ if defined? JRUBY_VERSION
       driver_jars
     end
 
-    if ENV['RUBYARCHDIR'] || ENV['BUILD_EXT_MAVEN'] == 'true'
+    if ENV['BUILD_EXT_MAVEN'] == 'true'
```
kares: at this point I am not sure why I did it (maybe its to be found in support branches' history) 
but you need to carefully evaluate all possibilities (some are in the comments) of the ext building being triggered.
e.g. with/without Bundler

kares: > There was a check on RUBYARCHDIR in the Rake task for the adapter jar which used to be set by the build script. This is not set anymore, so I removed the check.

we certainly weren't setting RUBYARCHDIR - believe it was simply from how the ENV ends up being when the ext is going to get build

kares:
while I understand your attempt to make it easier, I see the direction as being risky and redundant.
there's notes in BUILDING.md we do not need to solve a problem here of "auto-build"-ing the ext .jar
that could lead us a detection mechanism of trying to detect .java changes and if so building the ext ...
cause you do not want newbies getting confused when they make such change - better leave it explicit.

donv:
I wholeheartedly disagree.  The use case here is not for contributors to build the project (as was discussed), but for users to use ARJDBC as a github: source.  This used to work before `adapter_java.jar` was removed, and now it does not work.

The line in the gemspec is meant to be permanent and stay in the released version where it will be ignored since the `.git` directory is not present.

With this change ARJDBC is usable with Rails 5.1 + PG + Bundler.  Without the change it isn't.

kares:
@donv in that case I am sorry for going into conclusions too fast, the reason being is that ext .jar building used to work "before" as its implemented (and those changes that you touched upon here weren't needed). 

will double check whether `gem.extensions << 'Rakefile' if File.exist?('.git')` is a must.
... could you please post your Bundler/JRuby version, just to make sure we're on the same page here?

kares:
okey, so I definitely missed an important part while "back-porting" this feature:
https://github.com/jruby/activerecord-jdbc-adapter/blob/support/activerecord-jdbc-adapter.gemspec#L33-L49

... it is necessary to have `gem.extensions` added but we need to work-around it for the release.

kares:
ended up putting some of the suggested changes "in" ... but that is mostly due lack of time to resolve the ext building issue properly. am expecting we will still run into occasional issues with Bundler + `:git` source, hopefully not much.

 ideally **jar-dependencies** would solve the issues for us but I manage to run into some (known) issue even with latest *`TypeError' at org/jruby/RubyFile.java:916 - no implicit conversion of nil into String* ... that is with latest Bundler **1.16.1**, so we're back to local .jar resolution for now, reference commits : 

- https://github.com/jruby/activerecord-jdbc-adapter/commit/e5c9bcaf9e37cb3ffbad7308ca02f89b5f61cec2
- https://github.com/jruby/activerecord-jdbc-adapter/commit/3a643754499be059f31609a47174cb2488f83860
- https://github.com/jruby/activerecord-jdbc-adapter/commit/3043d1bff74445e6d135bf115dda418d81e39efe

donv:
I tested with JRuby 9.1.16.0 + Bundler 1.16.1 and it works well.  Thanks!

PR #882, opened 2018-04-16
Trying to fix postgres failure on ci

rdubya:


rdubya:
Didn't fix the ci issue, don't merge...

PR #888, opened 2018-04-18
Fixing static references to JRuby runtime state

headius:
This PR fixes the following two items:

* StringCache is now stored as an internal variable of ArJdbc mod.
* Loaded drivers cache is stored as an internal variable, I believe on ArJdbc, but unsure what "self" is in this context.

I request review from @enebo @kares @rdubya.

Specific concerns:

* Internal variables use instance variable logic, so they're a bit more overhead than a simple hash map.
* Obviously I have to look up ArJdbc module in the StringCache case. Both cases probably should store in same place using same mechanism but the loaded drivers cache has access to a "self" on which the load operation was called.

Unfixed at push time: RubyJdbcConnection has a static IRubyObject defaultConfig used for some JNDI logic. I was not clear what this field is or how to remove it. First instinct is to put it into the ARJDBC top-level module as an internal variable.

Related is the neighboring connectionFactory, which may hold runtime state.

### DIFF
--- a/src/java/arjdbc/ArJdbcModule.java
+++ b/src/java/arjdbc/ArJdbcModule.java
@@ -40,6 +40,7 @@ import org.jruby.exceptions.RaiseException;
 import org.jruby.runtime.Block;
 import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.builtin.IRubyObject;
+import org.jruby.runtime.builtin.InternalVariables;
 import org.jruby.util.ByteList;
 
 import arjdbc.jdbc.RubyJdbcConnection;
@@ -51,6 +52,8 @@ import static arjdbc.jdbc.RubyJdbcConnection.getJdbcConnection;
  * @author kares
  */
 public class ArJdbcModule {
+
+    private static final String ARJDBC_LOADED_DRIVERS = "arjdbc_loaded_drivers";
 
     public static RubyModule load(final Ruby runtime) {
         final RubyModule arJdbc = runtime.getOrCreateModule("ArJdbc");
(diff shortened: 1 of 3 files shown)


kares:
> StringCache is now stored as an internal variable of ArJdbc mod.

since its now double lookup it might be worth getting rid of it completely (measuring the impact).

> Loaded drivers cache is stored as an internal variable, I believe on ArJdbc, but unsure what "self" is in this context.

`WeakHashMap<Ruby, X>` was not holding a strong reference to runtime, what has been the issue here?
also I am not sure if its actually used where its mean to be -> establish conn loading drivers every time (e.g. `mysql_connection` which it used to have an impact on)

> RubyJdbcConnection has a static IRubyObject defaultConfig used for some JNDI logic.

that is just a fast compare (caching) mechanism to quickly decide the actual config used.


have been thinking about these leaks as I introduced them e.g. the `StringCache` did represent a measurable improvement at the time. con being non 'proper' multi-runtime friendliness but than most use-cases I can think of always load ext .class-es with JRuby's class-loader (which is per runtime). 

so it never really mattered ... and has proven "fine" since through the years (1.3 has StringCache) no one complained about memory leaking. which doesn't mean that its fine for 100% cases(?) but simply that 'complicating' these needs measuring to decide whether its actually still worth keeping.

enebo:
I do think we can get the global benefits across connections for stringCache by just making something which is constructed outside connection instances.  This is likely true for caching sites as well (although I really like the idea of rewriting bytecode from callMethod to indy via something like Mjlollnir, but that is a blue sky project near term).

headius:
> since its now double lookup it might be worth getting rid of it completely (measuring the impact).

I would expect two lookups to still be cheaper than a RubyString allocation (potentially with ByteList and byte[] as well), but I'll leave it up to you if you want to keep it.

> WeakHashMap<Ruby, X> was not holding a strong reference to runtime, what has been the issue here?

It's probably nothing. I wanted to be consistent in avoiding static references to runtime state. It didn't seem necessary for it to be global, but I admit I am not clear what it's saving us.

I also agree many risks are mitigated by having arjdbc loaded in a separate classloader, but we don't want to preclude other ways of deploying JRuby.

PR #894, opened 2018-05-05
Update the Regex to work with multline string

meetme2meat:
Its seem like the Index result returned can be a multiline string. Updating the regex to work with [multiline string](http://rubular.com/r/9ubFpmPtKD).


rdubya:
Thanks for the patch. Can you point me to what tests this fixes?

meetme2meat:
@rdubya ```ARCONN=jdbcpostgresql bin/test test/cases/schema_dumper_test.rb```

PR #901, opened 2018-07-16
Minor fixes for deprecation warnings, JRuby 9.2 and AR 5.2

dr-itz:


dr-itz:
more to come btw. (as time permits)...

enebo:
@dr-itz those were all pretty trivial to review but those three may have been better as 3 PRs since they were different logical changes.  So please keep contributing but don't worry about making smaller PRs.  We can handle that.
PS - I did not know why you picked a smaller usec value but @kares will jump in if he disagrees with it.

src/java/arjdbc/jdbc/RubyJdbcConnection.java
```diff
@@ -284,7 +284,7 @@ public static String formatTransactionIsolationLevel(final int level) {
     public static int mapTransactionIsolationLevel(final IRubyObject isolation) {
         final Object isolationString;
         if ( isolation instanceof RubySymbol ) {
-            isolationString = isolation.toString(); // RubySymbol.toString (interned)
```
kares: let's not change random lines if there isn't a reason ...
esp. since we might need to merge over all the way from 50-stable

kares: makes sense here - would have been nice to have it explained in a sep. commit

enebo: @kares yeah.  I also commented about separate commits too. As separate no doubt a better explanation would have happened too.

In this particular case in JRuby and here I prefer us to use asJavaString because toString is too varied in how it is used (e.g. one use case is for a debugging string).  asJavaString is more explicit that we want a string form (although this is a very minor change so I would have lived with it as toString).

PR #903, opened 2018-07-19
[postgres]: Improve handling of DB type "money"

dr-itz:
When using BigDecimal for money values, the standard handling with
setPGobjectParameter() does not work since it does a toString() on the
value. For a BigDecimal("123.45") this will produce "0.12345e3" which
PG doesn't like.

This registers a handler for PG type "money" which uses
statement.setBigDecimal() when a RubyBigDecimal is provided and falls
back to the standard handling setPGobjectParameter() for everything
else.

This fixes test errors:
* ajrdbc: test/db/postgresql/types_test.rb:608 in 'test_update_money'
* rails: test/cases/adapters/postgresql/money_test.rb:68:in `test_create_and_update_money'

### DIFF
--- a/src/java/arjdbc/postgresql/PostgreSQLRubyJdbcConnection.java
+++ b/src/java/arjdbc/postgresql/PostgreSQLRubyJdbcConnection.java
@@ -55,6 +55,7 @@ import org.joda.time.DateTimeZone;
 import org.jruby.*;
 import org.jruby.anno.JRubyMethod;
 import org.jruby.exceptions.RaiseException;
+import org.jruby.ext.bigdecimal.RubyBigDecimal;
 import org.jruby.javasupport.JavaUtil;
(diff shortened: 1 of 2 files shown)


kares:
makes sense - coudl you consider targeting this against 50-stable branch
than it would merge over to 51-stable (for new 50/51 release) and master as well

dr-itz:
ok, will do later today or tomorrow..

dr-itz:
Done. Cleanly cherry-picked for 50-stable, see #908

kares:
thanks, this will merge over eventually from 50-stable, from https://github.com/jruby/activerecord-jdbc-adapter/pull/903
now. we shall eventually close this one

dr-itz:
ok, so for future changes I'll target 50-stable unless it's AR 5.2 stuff only..

kares:
that would be really great, thanks again for the great stuff ...

PR #904, opened 2018-07-20
[test, postgres, sqlite] Fix explain with binds

dr-itz:
In Arel 9 used in AR 5.2, the way bind params work has changed:
  https://github.com/rails/arel/commit/db1bb4e9a728a437d16f8bdb48c3b772c3e4edb0
The values are now stored directly in the BindParam itself.

With this change, to_sql() in AR 5.2 now either accepts an Arel AST w/o
any binds or a SQL string with binds.

Change the tests to use the new syntax and test both ways. With this
in place fix PG explain with an Arel AST. Using to_sql() leads to no
binds at all and in turn makes PG complain. Instead use to_sql_and_bind()
which extracts the actual bind values from the Arel AST.

(Pretty sure plain AR5.2 on MRI is broken here for PG)

### DIFF
--- a/lib/arjdbc/postgresql/adapter.rb
+++ b/lib/arjdbc/postgresql/adapter.rb
@@ -339,8 +339,8 @@ module ArJdbc
     end
 
     def explain(arel, binds = [])
-      sql = "EXPLAIN #{to_sql(arel, binds)}"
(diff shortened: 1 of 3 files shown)


kares:
so this is AR 5.2 specific, right?

dr-itz:
just noticed on Travis: test_explain_with_binds fails for PG with PREPARED_STATEMENTS=false

Fix in a minute..

PR #905, opened 2018-07-20
[sqlite3] Fix connection.active?

dr-itz:
(just a line missing from recent import)


kares:
thanks

note-for-self: this is simple enough - should be picked to 50-stable/51-stable

dr-itz:
50-stable/51-stable should be fine w/o this

PR #906, opened 2018-07-21
[postgres] Fix NPE in internedTypeFor()

dr-itz:
attributeType() can return null. Then NPE is triggered when executing
rails52:test/cases/bind_parameter_test.rb:71

Check for null to prevent the exception. This turns the test above
from an error to a failure (missing deprecation warning).


dr-itz:
Ignore this...mixed it up...

PR #907, opened 2018-07-21
[postgres]: Support legacy binds

dr-itz:
Fixes:
* rails52:test/cases/adapter_test.rb#test_insert_update_delete_with_legacy

### DIFF
--- a/lib/arjdbc/postgresql/adapter.rb
+++ b/lib/arjdbc/postgresql/adapter.rb
@@ -325,7 +325,8 @@ module ArJdbc
     alias index_name_length max_identifier_length
 
     def exec_insert(sql, name, binds, pk = nil, sequence_name = nil)
-      val = super
+      binds = convert_legacy_binds_to_attributes(binds) if binds.first.is_a?(Array)
+      val = super(sql, name, binds, pk, sequence_name)
       if !use_insert_returning? && pk
         unless sequence_name
           table_ref = extract_table_ref_from_insert_sql(sql)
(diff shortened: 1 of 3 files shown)


dr-itz:
So...the NPE fix I'm still OK with, but..

Turns out sqlite3 has the same problem with legacy binds. So I wonder if the legacy binds
check should go to the generic code in ArJdbc::Abstract::DatabaseStatements?

Comments?

kares:
its broken - likely for all in 5.x ... recall I tried to fix it but than it ended up being reverted.
its likely tricky to get right - would need to dig up history to recall context/details on what went on ...

dr-itz:
Interesting...I'll do some digging to find out why it was problematic

(I just tried it in the generic code and it's simple enough...3 lines...but..)

kares:
believe I also wanted it to do less garbage - ideally not to render aby binds if SQL debug isn't enabled

enebo:
@dr-itz @kares so in 5.2 they realized they need to support legacy and re-added a test for it?  I don't see any legacy tests in 5.1.6 tests.

dr-itz:
I think it works in AR 5.0 and 5.1 (but lacking tests), the breakage was during 5.2 development. My proposed fix is here: #911

kares:
so with https://github.com/jruby/activerecord-jdbc-adapter/pull/911 this (postgres specific piece) is no longer needed, right?

dr-itz:
Yes...closing this for #911

PR #910, opened 2018-07-23
[52, mysql] Remove test_mysql_indexes

dr-itz:
In AR 5.2, INDEX_TYPES got removed, so just drop the test


dr-itz:
😃 ..and we have the first green build on Travis 😄

PR #911, opened 2018-07-23
[52] Support legacy binds for connection.(insert|update|delete)

dr-itz:
A legacy bind param is an array with two elements, column (often nil)
and the value. Turn those into proper bind params so they're handled
correctly.

Fixes
* rails52:test/cases/adapter_test.rb#test_insert_update_delete_with_legacy

Notes:
* Rails 5.0/5.1 don't have test for this, only Rails 5.2
* Tests were added in rails/rails@18abe37a for 5.2 along with a fix..
* ...for the breakage caused in rails/rails@213796f (after 5.1)

### DIFF
--- a/lib/arjdbc/abstract/database_statements.rb
+++ b/lib/arjdbc/abstract/database_statements.rb
@@ -42,10 +42,30 @@ module ArJdbc
       end
       alias :exec_delete :exec_update
 
+      # overridden to support legacy binds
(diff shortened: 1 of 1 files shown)


dr-itz:
@kares @enebo It's simple enough, fixes the test and it does not _seem_ to introduce any regressions, but I nevertheless suggest playing it safe here only only try that on master for Arjdbc-52...

Just let me know what you think...it's also easy to target 50-stable

kares:
LGTM

PR #914, opened 2018-07-24
[postgres] Exclude AR tests that use $1 instead of ? for binds

dr-itz:


enebo:
@dr-itz Is the goal of this to just exclude for now as it is a weird syntax?

dr-itz:
The $1,$2, etc. is basically the standard PG syntax, but the JDBC driver only handles the JDBC standard ? (and translates to $1 internally, but fails to handle $1 directly...which is disappointing) so this test can't work with JDBC. There's a few other excludes for the same reason already. In general, the Rails test don't use hard coded placeholders like this, with the exception of some adapter specific tests...

kares:
yep, we had skips due $X in place already ...

enebo:
@dr-itz ok makes sense.

PR #915, opened 2018-07-24
Skip test_explain_with_binds when PREPARED_STATEMENTS=false

dr-itz:
This version of explain is for Arel < 9 and cannot work w/o prepared
statements. The new-style Arel with bind values included in the AST
works fine. So just skip the test.

### DIFF
--- a/test/explain_support_test_methods.rb
+++ b/test/explain_support_test_methods.rb
@@ -27,6 +27,8 @@ module ExplainSupportTestMethods
   end
 
   def test_explain_with_binds
+    skip unless ActiveRecord::Base.connection.prepared_statements
+
     arel, binds = create_explain_arel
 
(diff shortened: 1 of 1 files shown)


enebo:
@dr-itz our current strategy with per branch releases is if a test is not applicable for the branch due to age we should just delete it.  So you can change this to remove that test.

dr-itz:
The test is OK for AR 5.2, but only with prepared statements...plain SQL string and array of binds. That can only work with prepared statements, otherwise the bind values are lost in exec_query...

(And I just fixed that test in #904 (almost that is), just never tested with PREPARED_STATEMENTS=false 😉 )

enebo:
@dr-itz Thanks for the explanation!

PR #918, opened 2018-07-30
[52, mysql] Update some rails test excludes

dr-itz:
* Exclude tests relying on mysql2 gem internals
* Rename an exclude where rails fixed a typo
* Delete an exclude for a test that no longer exists


dr-itz:
Done :)

PR #921, opened 2018-08-09
[52, sqlite] Trivial Rails test failure fixes

dr-itz:
Two commits with very trivial (almost too trivial to be worth the trouble) fixes for two Rails test failures.
(getting Rails SQLite tests green is so close)


enebo:
@dr-itz I just landed fixed for jruby 9.1 but this is great way to preserve things for 9.1 works with existing code.  We will revisit later if we want to version lock 5.2 support to 9.1.18+.

Soooo close to green!  Keep it up

Issue #899, opened 2018-06-08
Error with Rails 5.2


ckgagan:
Hi I was trying to upgrade to Rails 5.2

I pointed the gem to github master
`gem 'activerecord-jdbcpostgresql-adapter', github: 'jruby/activerecord-jdbc-adapter'`

The bundle install completed and installed all the gems.
But when I tried to run rails console, I got an error. Below is the stack trace of error

```
NameError: uninitialized constant ActiveRecord::ConnectionAdapters::PostgreSQL::ColumnDumper
                   load_missing_constant at /Users/ckgagan/.rbenv/versions/jruby-9.2.0.0/lib/ruby/gems/shared/gems/activesupport-5.2.0/lib/active_support/dependencies.rb:542
                           const_missing at /Users/ckgagan/.rbenv/versions/jruby-9.2.0.0/lib/ruby/gems/shared/gems/activesupport-5.2.0/lib/active_support/dependencies.rb:193
               <class:PostgreSQLAdapter> at /Users/ckgagan/.rbenv/versions/jruby-9.2.0.0/lib/ruby/gems/shared/bundler/gems/activerecord-jdbc-adapter-a33baa1c0e6a/lib/arjdbc/postgresql/adapter.rb:647
             <module:ConnectionAdapters> at /Users/ckgagan/.rbenv/versions/jruby-9.2.0.0/lib/ruby/gems/shared/bundler/gems/activerecord-jdbc-adapter-a33baa1c0e6a/lib/arjdbc/postgresql/adapter.rb:643
                                  <main> at /Users/ckgagan/.rbenv/versions/jruby-9.2.0.0/lib/ruby/gems/shared/bundler/gems/activerecord-jdbc-adapter-a33baa1c0e6a/lib/arjdbc/postgresql/adapter.rb:637
                                 require at org/jruby/RubyKernel.java:970
                         block in (root) at /Users/ckgagan/.rbenv/versions/jruby-9.2.0.0/lib/ruby/gems/shared/gems/activesupport-5.2.0/lib/active_support/dependencies.rb:1
                                 require at /Users/ckgagan/.rbenv/versions/jruby-9.2.0.0/lib/ruby/gems/shared/gems/activesupport-5.2.0/lib/active_support/dependencies.rb:283
[...]

PR #922, opened 2018-08-11
Pick the classpath of jruby from default path if not defined

prashantvithani:
Experienced the issue with java-10, sun.boot.class.path was not defined in default java environment variables. Noticed here that the only required path is of jruby.jar, which is usually defined as `<ruby-home>/lib/jruby.jar`

Reference: #899

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -237,7 +237,14 @@ if defined? JRUBY_VERSION
 
     classpath = []
     [ 'java.class.path', 'sun.boot.class.path' ].each do |key|
+      next if ENV_JAVA[key].empty?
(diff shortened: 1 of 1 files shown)


enebo:
Sorry I had written something up yesterday without pushing it :|  995be701d58f58a34bc0d1448c9cd54f9fcd9e48

This version assumes rbconfig libdir is main lib directory that contains jruby.jar.  This seems to work ok for rvm and ordinary tar.gz installs.  I don't think any of the other ruby managers put the jar in a different location.

Thanks for thinking about this problem.  Sorry I did not push this yesterday to make you spend brain cycles on it...

PR #927, opened 2018-08-21
[postgres] Fix postgresql_version parser for PG version 10 and higher

### LINKED ISSUE #926: [postgres-10] postgresql_version parser not compatible with PG version 10 & higher
According to the line [lib/arjdbc/postgresql/adapter.rb#L55](https://github.com/jruby/activerecord-jdbc-adapter/blob/master/lib/arjdbc/postgresql/adapter.rb#L55), The version string is expected to have exactly 3 digits `(9.6.10)`. According to the new [policy](https://www.postgresql.org/support/versioning/) by PostgreSQL, the versioning policy has changed and will contain only 2 digits e.g. `10.4`. This behaviour broke in many things depending on `postgresql_version` such as default behaviour of `insert_returning`.

prashantvithani:
Fix #926: From PostgreSQL version 10 and higher, the versioning policy
has changed. The updated method now captures whole version number
instead of fixed number of digits and returns integer that is compatible
to the depending methods and previous versions.

### DIFF
--- a/lib/arjdbc/postgresql/adapter.rb
+++ b/lib/arjdbc/postgresql/adapter.rb
@@ -52,7 +52,7 @@ module ArJdbc
       @postgresql_version ||=
         begin
           version = @connection.database_product
-          if version.match /PostgreSQL ((\d+\.)+\d+)/
+          if version.match /PostgreSQL (\d+.*)/
             version_numbers = $1.split('.').map(&:to_i)
             # PostgreSQL version representation does not have more than 3 digits
             return 0 if version_numbers.length > 3
--- a/lib/arjdbc/postgresql/adapter.rb
+++ b/lib/arjdbc/postgresql/adapter.rb
@@ -52,8 +52,15 @@ module ArJdbc
       @postgresql_version ||=
         begin
           version = @connection.database_product
-          if version =~ /PostgreSQL (\d+)\.(\d+)\.(\d+)/
-            ($1.to_i * 10000) + ($2.to_i * 100) + $3.to_i
+          if version.match /PostgreSQL ((\d+\.)+\d+)/
+            version_numbers = $1.split('.').map(&:to_i)
+            # PostgreSQL version representation does not have more than 3 digits
+            return 0 if version_numbers.length > 3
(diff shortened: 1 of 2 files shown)


lib/arjdbc/postgresql/adapter.rb
```diff
@@ -52,8 +52,15 @@ def postgresql_version
       @postgresql_version ||=
         begin
           version = @connection.database_product
-          if version =~ /PostgreSQL (\d+)\.(\d+)\.(\d+)/
-            ($1.to_i * 10000) + ($2.to_i * 100) + $3.to_i
+          if version.match /PostgreSQL ((\d+\.)+\d+)/
+            version_numbers = $1.split('.').map(&:to_i)
```
kares: match does not use `$1`, thus this seems self invented (not AR core port). 
maybe its better to have what Rails is using or if its different at least a unit test for the 'compicated' versioning part?
also did this get fixed in AR 5.0 or 5.1?

lib/arjdbc/postgresql/adapter.rb
```diff
@@ -52,8 +52,15 @@ def postgresql_version
...
+            # PostgreSQL version representation does not have more than 3 digits
+            return 0 if version_numbers.length > 3
```
dr-itz: There are servers with 4 version parts, e.g. some EnterpriseDB distributions. The JDBC driver just ignores the 4th part, see `pgjdbc/src/main/java/org/postgresql/core/ServerVersion.java`. It's referenced with a bug: https://github.com/pgjdbc/pgjdbc/issues/667. ARJDBC should ignore a 4th part as well (if there is one).

dr-itz:
@prashantvithani Thanks for this. I took the liberty and added the 4-part version thing I mentioned and rebased for 50-stable so we can have this for 50.x, 51,x, 52.x, see #932.

prashantvithani:
@dr-itz I wasn't aware of the 4-part versioning in EnterpriseDB. Thanks for pointing it out and fixing it.

enebo:
@prashantvithani closing in favor of the newer PR from @dr-itz

PR #929, opened 2018-08-22
Travis: use postgresql-9.4

dr-itz:
By default Travis uses PG 9.2. But it seems the client is actually newer
than that. One of the test errors is the server complaining about
"lock_timeout" which is a 9.3+ thing. Using a newer server version
fixes the problem. 9.2 is EOL anyway and 9.3 will be in a few weeks,
so go for 9.4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,8 @@ rvm:
   - jruby-9.1.16.0
 jdk:
   - openjdk8
+addons:
+  postgresql: "9.4"
 env:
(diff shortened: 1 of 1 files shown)


enebo:
@dr-itz I think using 5.2+ for PG 9.3 while 5.0 and 5.1 using PG 9.2 for the moment is a reasonable way of balancing out testing.  I added this note because I almost asked you to rebase back to 50-stable.

PR #930, opened 2018-08-22
[52, sqlite] Fix rake tasks

dr-itz:
Rails' SQLite tasks uses Kernel.system() with "out:" which JRuby does
not yet support. Import the task module, use backticks with output
redirection instead of Kernel.system(). Also exclude a test that tests
the usage of Kernel.system()

Fixes:
* rails52/test/cases/tasks/sqlite_rake_test.rb


kares:
not sure this is a good enough fix, its now `LOAD_PATH`dependent - having the same .rb file available ...

dr-itz:
Yeah, that won't do. Just pushed another version that should be independent of LOAD_PATH, but it's ugly and makes me feel dirty

PR #934, opened 2018-08-25
[50-52, MariaDB] Fall back to 'localhost' on empty host string

dr-itz:
The MariaDB driver doesn't like an empty host in the URL. E.g. this
works with the MySQL driver but doesn't with the MariaDB driver:
  jdbc:mysql:///activerecord_unittest
Simply fall back to 'localhost' if host is empty for MariaDB driver.
While at it, make #mysql_connection() a bit easier to read.


dr-itz:
This makes the Rails tests run for MariaDB on Travis.
(Only noticed the failure on Travis since I had my config.yml pointing to a different host...)

PR #941, opened 2018-08-30
Add location of jruby.jar to classpath from rbconfig for java 9+

prashantvithani:
Both Rakefile and `compile.rake` define the task
`lib/arjdbc/jdbc/adapter_java.jar`. So, while calling `rake` to compile
the gem, the task gets invoked twice - First from Rakefile followed by
the task defined in `compile.rake`. The classpath issue for Java 9+ was
fixed in Rakefile but not in `rakelib/compile.rake`.

### DIFF
--- a/rakelib/compile.rake
+++ b/rakelib/compile.rake
@@ -1,6 +1,7 @@
 jar_file = File.join(*%w(lib arjdbc jdbc adapter_java.jar))
 begin
   require 'ant'
+  require 'rbconfig'
   directory classes = "pkg/classes"
   CLEAN << classes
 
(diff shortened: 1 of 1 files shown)


enebo:
@prashantvithani hey can you rebase this againt 50-stable branch. I think all 5.x releases should include this.

prashantvithani:
@enebo Done!

enebo:
@prashantvithani thanks.  hopefully we got all the cases now.

prashantvithani:
@enebo Thanks for merging the changes. It looks like we have covered all cases. But, I still don't understand the purpose of defining the task `:jar => lib/arjdbc/jdbc/adapter_java.jar` twice. Could you please explain what is the objective of the task defined in `rakelib/compile.rake`?

enebo:
@prashantvithani  Thanks for bringing this up again.  I believe @kares in last year had moved that code into Rakefile and forgot to delete it in compile.rake?  @kares I assume you prefer what is in Rakefile since you put it there last?

kares:
hmm, I am not sure we actually use (need) compiler.rake, does these change make a difference at all?
... might work for local dev (as rakelib/*.rake gets auto-imported) but I am not sure whether it would work with (bundler) gem git: ... checkouts. might have moved it into *Rakefile* due issues, not sure, we shall pbly just delete the second one instead of back-porting the same changes as we did here.

enebo:
@kares please do.  The code from rakelib was put in 2011 so it was the original snippet.  I don't care which one stays but I know less about this than you do...

kares:
yeah, makes sense - also removes `require 'ant'` + will get into this - keeping some manifest pieces

PR #942, opened 2018-08-30
[postgres] Use pgjdbc "autosave" option for hot compat

dr-itz:
The pgjdbc driver internally caches parsed statements (the most heavy
part of prepareStatement()). The hot_compatibility.rb test for INSERT
fails because the internal cache is used, not the ARJDBC one (only used
for queries).

Setting the option "autosave = ' conservative'", the driver handles
the cached plan changes internally: it creates a savepoint before the
query, when the problem is hit, it rolls back to the savepoint, clears
the cache and retries. This works for all statements and even when
inside a transaction. No more special handling required.

Fixes:
* rails/test/cases/hot_compatibility_test.rb

### DIFF
--- a/lib/arjdbc/postgresql/adapter.rb
+++ b/lib/arjdbc/postgresql/adapter.rb
@@ -689,21 +689,6 @@ module ActiveRecord::ConnectionAdapters
       TableDefinition.new(*args)
     end
 
-    def exec_query(sql, name = nil, binds = [], prepare: false)
-      super
-    rescue ActiveRecord::StatementInvalid => e
-      raise unless e.cause.message.include?('cached plan must not change result type'.freeze)
-
-      if open_transactions > 0
-        # In a transaction, have to fail it - See AR code for details
-        raise ActiveRecord::PreparedStatementCacheExpired.new(e.cause.message)
-      else
-        # Not in a transaction, clear the prepared statement and try again
-        delete_cached_statement(sql)
-        retry
-      end
-    end
-
     public :sql_for_insert
 
     def schema_creation # :nodoc:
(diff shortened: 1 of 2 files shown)


dr-itz:
Note: this could easily target 50-stable as well, but I'm not sure it's a good idea since 50.x and 51.x have so many more users (according to rubygems.org downloads) than 52.x. Dunno. I can always rebase for 50-stable or 51-stable, just say so ;)

kares:
looking good, the `exec_query` removal looks awesome, there likely isn't any similar code in core, right?
would certainly vote for moving this to all 5.x ... if @rdubya does not mind?

dr-itz:
AR with pg gem basically does the same in `exec_cache` that arjdbc does in `exec_query`. Since pgjdbc does caching internally, we would need to do the check for `exec_insert`, `exec_delete` and `exec_update` as well. Or disable the pgjdbc cache, but that would be bad for performance.

pgjdbc's internal caching does give a performance boost even for insert/update that are not cached in arjdbc. This even works with automatic server side prepared statements (according to the docs). And the `autosave` has the advantage that it works for transactions as well, so arjdbc is better off 😄

I'll rebase to 50-stable once @rdubya gives his ok.

rdubya:
Looks good to me, definitely like that it removes the need to string match on the exception's message. I think it would be good to get into 50 and 51, the only side effect that anybody should notice is if they are dependent on that exception being thrown when the cache is stale right? I can't think of a use case for doing that, so we're probably good to put it everywhere.

dr-itz:
@rdubya thx for the feedback
@kares it's now rebased to 50-stable

PR #945, opened 2018-08-30
[50-52, mysql] Fix CI encoding problems

### LINKED ISSUE #879: excluded MySQL tests failing due UTF-8 issues on CI
these seems to only fail on CI and only while executing with the Rails tests.
they do work locally and when test pieces are transplanted to AR-JDBC's suite.

have investigated the difference between test DB setups as well as CI's MySQL configuration.
the outcome is just weird and not clear why they're would be failing.
its to be noted that MySQL has some gotchas with UTF-8 an `encoding: utf8` does not mean Unicode,
but I have also tried **utf8mb4** with the same outcome and that does not explain why they pass locally.

sample test (failure) details : 

- https://travis-ci.org/jruby/activerecord-jdbc-adapter/jobs/360275647#L691
- https://travis-ci.org/jruby/activerecord-jdbc-adapter/jobs/360275647#L735
- https://travis-ci.org/jruby/activerecord-jdbc-adapter/jobs/360275647#L801
- https://travis-ci.org/jruby/activerecord-jdbc-adapter/jobs/360275647#L856
- https://travis-ci.org/jruby/activerecord-jdbc-adapter/jobs/360275647#L921

dr-itz:
Encoding of the test DBs was wrong...

Fixes #879

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,8 @@ before_script:
     mysql -e "grant all privileges on activerecord_unittest.* to rails@localhost;" && \
     mysql -e "grant all privileges on activerecord_unittest2.* to rails@localhost;" && \
     mysql -e "grant all privileges on inexistent_activerecord_unittest.* to rails@localhost;" && \
-    mysql -e "CREATE DATABASE activerecord_unittest;" && \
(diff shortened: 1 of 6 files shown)


kares:
very nice, thanks!

PR #949, opened 2018-09-05
[50-52, postgres] Add missing supports_x? methods

dr-itz:
This mostly makes sure that the tests for these features are
actually executed.

### DIFF
--- a/lib/arjdbc/postgresql/adapter.rb
+++ b/lib/arjdbc/postgresql/adapter.rb
@@ -224,6 +224,8 @@ module ArJdbc
 
     def supports_ddl_transactions?; true end
 
+    def supports_advisory_locks?; true end
+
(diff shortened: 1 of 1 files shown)


enebo:
@dr-itz I have only seen 50-stable test come back but this causes fails with PS of postgresql because bulk_alter_table method is not defined.

dr-itz:
`supports_bulk_alter?` is only supported on AR5.2, the rest should be ok

enebo:
@dr-itz  no harm no foul.   I can just remove that from 50 and 51

enebo:
@dr-itz done.  I somehow added whitespace in 52 and master after merge...so a little harm but I inflicted it :)

dr-itz:
@enebo thanks

PR #951, opened 2018-09-06
Fixes #891. Modifies super-call for log method in AbstractAdapter.

### LINKED ISSUE #891: jdbc_connection method hides ArgumentErrors thrown from 4-arg adapter_class constructors
# Summary

[This rescue](https://github.com/jruby/activerecord-jdbc-adapter/blob/v50.0/lib/arjdbc/jdbc/connection_methods.rb#L16) hides ArgumentErrors thrown from 4-argument adapter_class constructors

# Environment

* activerecord-jdbc-adapter 50.0

# Steps

Attempt to initialize an adapter with a 4-argument constructor that throws an ArgumentError (in my case it was activerecord-postgis-adapter)

# Expected

The exception thrown from the constructor

# Actual

"Wrong number of arguments (3 for 4)" from [this line](https://github.com/jruby/activerecord-jdbc-adapter/blob/v50.0/lib/arjdbc/jdbc/connection_methods.rb#L17)

patrickj0607:
Removes the ArgumentError by providing the 4 required arguments explicitly.

### DIFF
--- a/lib/arjdbc/abstract/core.rb
+++ b/lib/arjdbc/abstract/core.rb
@@ -67,7 +67,7 @@ module ArJdbc
         if binds.any? && (type_casted_binds.nil? || type_casted_binds.empty?)
           type_casted_binds = ->{ extract_raw_bind_values(binds) }
         end
-        super
+        super(sql, name, binds, statement_name)
       end
     end
   end


dr-itz:
That can't be right. It will breaks logging of bind params again.

Rails 5-0-stable (and later) has that defined as 

```
def log(sql, name = "SQL", binds = [], type_casted_binds = [], statement_name = nil)
```

dr-itz:
Ok, got an explanation for this. The signature of the `log` method was changed in AR 5.0.3, the `type_casted_binds` got added. So it look like you're using Rails 5.0.2 or older. You should upgrade to 5.0.7.

enebo:
@dr-itz @kares should we update the minimum version of 50-stable to 5.0.3 or maybe just go to 5.0.7+?

@patrickj0607 Let us know if you are on an earlier version or we have another mystery to work out.

patrickj0607:
@dr-itz @enebo You were right, I was on rails 5.0.1. After upgrading to 5.0.7, I could run it without any problem. As that makes this PR obsolete, I will close it. Thanks for all your support!

dr-itz:
@patrickj0607 thanks for testing

@enebo @kares I think raising the minimum version would be good. I did a few CI runs to see how the different Rails versions behave:
* 5.0.3: https://travis-ci.org/dr-itz/activerecord-jdbc-adapter/builds/426164565
* 5.0.4: https://travis-ci.org/dr-itz/activerecord-jdbc-adapter/builds/426405319
* 5.0.5: https://travis-ci.org/dr-itz/activerecord-jdbc-adapter/builds/426413326
* 5.0.6: https://travis-ci.org/dr-itz/activerecord-jdbc-adapter/builds/426422408
* 5.0.7: https://travis-ci.org/jruby/activerecord-jdbc-adapter/builds/425442851

So 5.0.3-5.0.6 basically look more less the same, 5.0.7 has three "log" related test failures with PS less. I guess 5.0.3 should be OK as a minimum. On the other hand 5.0.7 works best since it's what we have been testing with for some time....dunno

rdubya:
@dr-itz yeah that makes sense, hate it when I get stuck in a dependency mess

kares:
should be done `">= 5.0.3"` for *50-stable* 
we really should be conservative here (not forcing latest AR), only doing these if we need to (like in this case, as you guys figured out - older were broken and we had users complaining). lots of users might get into a bad state when they need to ugrade several gems (rails) due another gem -> although it should be a no brainer with patch releases I am not sure its actually a reality in 5.x

rdubya:
@kares What's the difference between `'~> 5.0.0', '>= 5.0.3'` and `'~>5.0.3'` ? I think I've seen it done both ways but could never figure out the difference between the two.

kares:
duno - expect it to be the same as those 2 combined, 
wanted to keep `'~> ...'` and add so merging is clear - do not end up resolving a conflict as `'~> 5.2.3'`

rdubya:
ah, yeah that makes sense

PR #952, opened 2018-09-12
[postgres] Update version regex matcher to ignore anyting after version number

prashantvithani:
PPA repositories (of Ubuntu) appends its own version string after 'PostgreSQL <version-number>' i.e. it becomes 'PostgreSQL <version-number> <version-string-by-PPA>'. E.g: PostgreSQL 10.4 (Ubuntu 10.4-2.pgdg16.04+1)

### DIFF
--- a/lib/arjdbc/postgresql/adapter.rb
+++ b/lib/arjdbc/postgresql/adapter.rb
@@ -52,7 +52,7 @@ module ArJdbc
       @postgresql_version ||=
         begin
           version = @connection.database_product
-          if version.match /PostgreSQL (\d+.*)/
+          if version.match /PostgreSQL\s*([\d\.]*\d).*/
             version_numbers = $1.split('.').map(&:to_i)
             # PostgreSQL version representation does not have more than 4 digits
             return 0 if version_numbers.length > 4
--- a/test/db/postgresql/version_test.rb
+++ b/test/db/postgresql/version_test.rb
@@ -23,6 +23,10 @@ class VersionTest < Test::Unit::TestCase
     assert_equal connection_stub('9'), 90000
   end
 
+  def test_pg_version_with_os_version
+    assert_equal connection_stub('10.4 (Ubuntu 10.4-2.pgdg16.04+1)'), 100400
+  end
+
   private
 
   def connection_stub(version_string)


kares:
is this the same logic as **pg** does? (since its been changing frequently)
there's also a way of changing the "PostgreSQL" name, right? so why have it there?
// cc @rdubya

prashantvithani:
@kares thanks for pointing this out. I'm not sure if we can use the functions used by `pg` in https://www.rubydoc.info/gems/pg/PG%2FConnection:server_version as it's from C-library as mentioned in the link below.

`pg` gem uses `PQserverVersion`, which returns the version integer directly. 
Reference: https://www.postgresql.org/docs/10/static/libpq-status.html

According to what mentioned in the doc from above link (`libpq-status`), the [current implementation](https://github.com/jruby/activerecord-jdbc-adapter/blob/master/lib/arjdbc/postgresql/adapter.rb#L51) of converting version to integer for postgres 10 (2-part version system) is wrong. It was correct for 3-part version system though.

> Applications might use this function to determine the version of the database server they are connected to. The result is formed by multiplying the server's major version number by 10000 and adding the minor version number. For example, version 10.1 will be returned as 100001, and version 11.0 will be returned as 110000. Zero is returned if the connection is bad.
>
> Prior to major version 10, PostgreSQL used three-part version numbers in which the first two parts together represented the major version. For those versions, PQserverVersion uses two digits for each part; for example version 9.1.5 will be returned as 90105, and version 9.2.0 will be returned as 90200.

I will change this. But, still not sure if pgjdbc driver provides the facility to fetch the pg version as integer. Also, I've kept "PostgreSQL" just for the consistency purpose as it was there already. It might help in ensuring that the version string returned by `databse_product` starts with `PostgreSQL` only as it's hardcoded in pgjdbc [PgDatabaseMetaData.java#L138](https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/jdbc/PgDatabaseMetaData.java#L138)

dr-itz:
Eeew...missed that part with the >= 10 versions last time too, even tough I had a look at the code in pgjdbc. The libpq implementation is here:
https://github.com/postgres/postgres/blob/master/src/interfaces/libpq/fe-exec.c#L1055
It does string parsing too. It seems that it completely ignores the "PostgreSQL".

It's too bad actually...your implementation with "zip" was elegant :(

The JDBC does unfortunately not have a public API to get the version as an int...

kares:
so I've decided to setup something slightly different at https://github.com/jruby/activerecord-jdbc-adapter/commit/08bfc74ec5e8fa79b27ddb95065d5d47eb950bf5 ... should be very permissive and just take the first version looking number it finds.

sorry for skipping a PR but I'd like to do a release soon-ish. 
hopefully, @dr-itz and @rdubya approves the 'new' change.

dr-itz:
@kares problem is, as @prashantvithani figured out in the doc, "10.1 will be returned as 100001" in libpq, so 10.1 should be handled like 10.0.1...and there is some versions like 9.6devel that should be handled like 9.6.0. libpq has different cases for 1/2/3 parts with special handling for major >= 10 https://github.com/postgres/postgres/blob/master/src/interfaces/libpq/fe-exec.c#L1055-L1089
so I think we should just do the same...

(really too bad I missed that the last time)

dr-itz:
@kares, @prashantvithani See #953. Ain't pretty, but...

PR #953, opened 2018-09-13
[50-52, postgres] Fix PG version parsing for PG >= 10

dr-itz:
Do the same as libpq, see
https://github.com/postgres/postgres/blob/master/src/interfaces/libpq/fe-exec.c


prashantvithani:
@dr-itz Thanks for making the change! Hopefully, we have covered all the cases now!

kares:
Great Daniel, this is what I was hoping to get as I asked "is this the same logic as pg does?"
... code looks fine and I trust you did the best to match PG's C so will push this into the release.

Issue #955, opened 2018-10-04
ArgumentError (wrong number of arguments (4 for 3)) using JDBC


dgpokl:
### Summary
I am using JDBC to connect to an AS/400 database and I'm seeing a fatal error as soon as I define a model and call `establish_connection`. The error is raised at `lib/arjdbc/abstract/core.rb:12` which seem to be related to the 4 vs 3 argument constructors on database adapters.

As you can see in the stacktrace below:
[lib/arjdbc/jdbc/connection_methods.rb:8](https://github.com/jruby/activerecord-jdbc-adapter/blob/master/lib/arjdbc/jdbc/connection_methods.rb#L8) calls the adapter's constructor using 4 arguments.
[lib/arjdbc/jdbc/callbacks.rb:12](https://github.com/jruby/activerecord-jdbc-adapter/blob/master/lib/arjdbc/jdbc/callbacks.rb#L12) calls `super` **without arguments,** passing the 4 arguments through to...
[lib/arjdbc/abstract/core.rb:12](https://github.com/jruby/activerecord-jdbc-adapter/blob/master/lib/arjdbc/abstract/core.rb#L10) which has a 3-argument constructor. Thus I get the error message:
`ArgumentError (wrong number of arguments (4 for 3))`

I read with interest the comments on (especially) #896, on #891, and on the last 2 commits [here](https://github.com/jruby/activerecord-jdbc-adapter/commits/master/lib/arjdbc/jdbc/connection_methods.rb) and it seems like the goal now is to only use 4-argument constructors throughout? So if that's true, **don't the adapters defined in this gem need to be updated accordingly, including this one?** Or (and I acknowledge this is a very real possibility) am I just using it wrong? Or is this fixed in master, which I haven't yet gotten to successfully install? Any help or guidance would be **greatly** appreciated.

To try to understand what was going on since this code in callbacks.rb is a mix-in, I made a couple quick edits to log the names of the classes involved at callbacks.rb:12 and confirmed that the current class is `ActiveRecord::ConnectionAdapters::JdbcAdapter` and the superclass it's calling super to reach is `ActiveRecord::ConnectionAdapters::AbstractAdapter`.

### Steps to reproduce
Here's a basic setup though that might suffice to demonstrate the issue. Note: I tried this leaving out steps 2-3 and it looks like you can get it to throw the same error without having the jar file even in place, so you may be able to skip them too. And it seems to me like that might mean it's not just people with super similar setups to me.
[...]

PR #957, opened 2018-10-30
[db2] Add DB2 and AS400 support  for AR52

rsov:
Fixes https://github.com/jruby/activerecord-jdbc-adapter/issues/955 and adds basic support for DB2/AS400

Had to add ArJcbc::AR52 to the JdbcColumn class to initialize adapter.

### DIFF
--- a/lib/arjdbc/db2/adapter.rb
+++ b/lib/arjdbc/db2/adapter.rb
@@ -30,6 +30,65 @@ module ArJdbc
   # @note This adapter doesn't support explain `config.active_record.auto_explain_threshold_in_seconds` should be commented (Rails < 4.0)
   module DB2
 
+
+    module ActiveRecord::ConnectionAdapters
+
+      remove_const(:DB2Adapter) if const_defined?(:DB2Adapter)
+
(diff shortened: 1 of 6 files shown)


enebo:
@rsov in future commits/PRs you can remove any version checks or code for less than AR 5.2 unless you explicitly want to go back to 5.0+ but personally I think we should push forward from 5.2.  Our plan moving forward is to just only have the exact code we need for each versioned branch and stop maintaining version checks.

rsov:
@edsono Thank you for the quick merge. 

I was thinking a similar thing. We could set up folders for each version of AR. That way we can keep existing support but also have clear code separation

enebo:
@rsov we handle these separate versions just using 50-stable, 51-stable, 52-stable (and soon to have 60-stable). Each one only has changes necessary for the version they represent.  So after I merged your PR I merged 52-stable with master (which is where current 6.0 work is being done.  Thus far the deltas of the other main adapters has been reasonably small across version changes.  The main icky part of this strategy is quite a few merge commits...

rsov:
@enebo Makes sense. Sounds like you guys figured out a good solution

PR #959, opened 2018-10-31
Fixed for AR52

rsov:
- Log in rails dev server console now shows where query was executed from
- `User.exists? 1` won't fail due to ID not being passed

### DIFF
--- a/lib/arjdbc/abstract/core.rb
+++ b/lib/arjdbc/abstract/core.rb
@@ -22,7 +22,7 @@ module ArJdbc
 
         connection.configure_connection # will call us (maybe)
       end
-      
+
       # Retrieve the raw `java.sql.Connection` object.
       # The unwrap parameter is useful if an attempt to unwrap a pooled (JNDI)
       # connection should be made - to really return the 'native' JDBC object.
@@ -71,6 +71,21 @@ module ArJdbc
         end
         super
       end
+    end
+  end
+end
+
+
+module ActiveRecord
+  class LogSubscriber
+
+    JDBC_GEM_ROOT = File.expand_path("../../../..", __FILE__) + "/"
+
+    # Remove this gem from log trace so that query shows where it was called in application
+    def ignored_callstack(path)
+      path.start_with?(JDBC_GEM_ROOT) ||
+      path.start_with?(RAILS_GEM_ROOT) ||
+      path.start_with?(RbConfig::CONFIG["rubylibdir"])
     end
   end
 end
--- a/lib/arjdbc/abstract/core.rb
+++ b/lib/arjdbc/abstract/core.rb
@@ -73,19 +73,14 @@ module ArJdbc
       end
     end
   end
-end
 
-
-module ActiveRecord
-  class LogSubscriber
-
+  module LogSubscriber
     JDBC_GEM_ROOT = File.expand_path("../../../..", __FILE__) + "/"
 
     # Remove this gem from log trace so that query shows where it was called in application
     def ignored_callstack(path)
-      path.start_with?(JDBC_GEM_ROOT) ||
-      path.start_with?(RAILS_GEM_ROOT) ||
-      path.start_with?(RbConfig::CONFIG["rubylibdir"])
+      super || path.start_with?(JDBC_GEM_ROOT)
     end
   end
+  ActiveRecord::LogSubscriber.prepend(ArJdbc::LogSubscriber)
 end
--- a/lib/arjdbc/jdbc/adapter.rb
+++ b/lib/arjdbc/jdbc/adapter.rb
@@ -250,12 +250,6 @@ module ActiveRecord
         end
       end
 
-      # @private
-      # @override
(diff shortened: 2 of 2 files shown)


lib/arjdbc/abstract/core.rb
```diff
@@ -74,3 +74,18 @@ def log(sql, name = "SQL", binds = [], type_casted_binds = [], statement_name =
...
 end
+
+
+module ActiveRecord
+  class LogSubscriber
+
+    JDBC_GEM_ROOT = File.expand_path("../../../..", __FILE__) + "/"
+
+    # Remove this gem from log trace so that query shows where it was called in application
+    def ignored_callstack(path)
```
rdubya: I'm not sure I follow what this is doing 100%, but won't it make it harder for us to debug issues when the problem lies in the gem/rails?

rdubya:
What database are you testing with?

lib/arjdbc/jdbc/adapter.rb
```diff
@@ -250,12 +250,6 @@ def exec_query_raw(sql, name = 'SQL', binds = [], &block)
         end
       end
 
-      # @private
-      # @override
-      def select_rows(sql, name = nil, binds = [])
```
kares: also, I do not understand the removal from the desc or comit message: "Fixed issue with User.exists? 1 call"
could you elaborate? are you trying to fit the base adapter (un-used within core) with the DB2 adapter as is?

lib/arjdbc/abstract/core.rb
rsov: Actually the opposite. All that method does is figure out which line executed the AR call and logs it into rails sever log if `ActiveRecord::Base.verbose_query_logs`. See [source](https://github.com/rails/rails/blob/fc5dd0b85189811062c85520fd70de8389b55aeb/activerecord/lib/active_record/log_subscriber.rb#L102). It already removes AR gem from stack trace. 

Without the fix (same line every time)

```
User Load (7.0ms)  SELECT  * FROM USERS WHERE id = ? LIMIT ?  [["id", "1"], ["LIMIT", 1]]
  ↳ /home/ruslan/rails/rsov/activerecord-jdbc-adapter/lib/arjdbc/abstract/core.rb:72
```

Expected (line changes based where it was called from)
```
User Load (7.0ms)  SELECT  * FROM USERS WHERE id = ? LIMIT ?  [["id", "1"], ["LIMIT", 1]]
  ↳ app/controllers/application_controller.rb:107


Member Load (59.2ms)  SELECT * FROM MEMBERS WHERE id = ? LIMIT ?  [["id", 1], ["LIMIT", 1]]
  ↳ app/views/layouts/_details.html.erb:6
```

lib/arjdbc/jdbc/adapter.rb
rsov: This was a tricky one.

When I execute query `User.find(1)` it does what you expect, executes this query: 
`SELECT  * FROM USERS WHERE id = ? LIMIT ?  [["id", "1"], ["LIMIT", 1]]`

However just doing `User.exists?(1)` does this:
```
ActiveRecord::StatementInvalid: ActiveRecord::JDBCError: [SQL0313] Number of host variables not valid.: SELECT  1 AS one FROM USER WHERE id = ? LIMIT ?
```

It lost the binds variables (`[["id", "1"], ["LIMIT", 1]]`)

Originally I though, yeah, another DB2 bug. But DB2 does not have `select_rows`. And if you print the binds variables from `select_rows`, it's empty

However the binds do exists. The first variable (`sql`) is changed to `arel` which contains building query information. Then AR52 [select_rows](https://github.com/rails/rails/blob/fc5dd0b85189811062c85520fd70de8389b55aeb/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb#L80) calls  [select_all](https://github.com/rails/rails/blob/fc5dd0b85189811062c85520fd70de8389b55aeb/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb#L46) which parses out the binds from the arel query.

Removing `select_rows` from JDBC fixed the issue

dr-itz: ah yes, the AREL 5.2 change where the bind values are in the AST and need to be extracted. Nice catch.

rsov:
@rdubya Currently testing on DB2/AS400

lib/arjdbc/abstract/core.rb
kares: nice, thanks @rsov ... think it all makes sense.
@rdubya if you do not have anything else I'm fine with merging - feel free to go forward.

rdubya: Can you update it to create a module that we prepend to the class that then calls `super`? That way we don't have to come back and update this if they ever change what they are filtering out

rsov: @rdubya Done

rdubya: Cool. Thanks

PR #966, opened 2018-12-13
Changed params of type_to_sql to match the ones of AR5.2

rsov:
Currently migrations are failing on AR5.2 because the `type_to_sql` was changed from 4 parameters to 2 parameters. 

Changed the method in db2 adapter to match this

### DIFF
--- a/lib/arjdbc/db2/adapter.rb
+++ b/lib/arjdbc/db2/adapter.rb
@@ -487,9 +487,9 @@ module ArJdbc
       types
     end
 
-    def type_to_sql(type, limit = nil, precision = nil, scale = nil)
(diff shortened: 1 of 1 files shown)


enebo:
@rsov was that done in a point release of 5.2.x or have we always been broken?

kares:
@enebo this is DB2 adapter only - which we did not release (or declare that we will support) for AR 5.x

enebo:
@kares ah ok.  I should have read the filename :)

 We did not delete the DB2 files and someone seems to be hacking so I think we should allow this.  Whether we decide to officially support a release depends on whether someone decides they want to commit to supporting this with us.

PR #967, opened 2018-12-31
[DB2] Refactor DB2 quote method for AR5.1 and up

rsov:
In [Active Record 5.1 ](https://github.com/rails/rails/commit/e646bad5b7c462187fc75c11d5c06e43759022d8#diff-15fcde66de1c8e6b1c64d6ff32f6f223L8) and up secondary parameter `column` has been deprecated and  dropped. 

There are a lot of checks of `column` parameter inside `DB2 quote` method, those checks have been removed. 

Also checks like `when nil then 'NULL'` are done in ActiveRecord's `_quote` method. Removed them and call super to handle it


kares:
yet, these should end up to have to go somewhere? 
are you seeing the same no of failures before and after this patch?

rsov:
Take a look at [AR quote](https://github.com/rails/rails/blob/e646bad5b7c462187fc75c11d5c06e43759022d8/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb#L8) that calls [_quote](https://github.com/rails/rails/blob/e646bad5b7c462187fc75c11d5c06e43759022d8/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb#L153)

_quote has the case statements that were inside DB2 quote method. Any overrides that we put in, like `quote_time` should still get triggered form AR.

The issue I was having comes from DB2 quote calling `super` and passing it 2 parameters. `ArgumentError: wrong number of arguments (2 for 1)` [here](https://github.com/jruby/activerecord-jdbc-adapter/blob/5eb2d922a50d434d793f0160702aceb0823c645a/lib/arjdbc/db2/adapter.rb#L454) and [here](https://github.com/jruby/activerecord-jdbc-adapter/blob/5eb2d922a50d434d793f0160702aceb0823c645a/lib/arjdbc/db2/adapter.rb#L456)

I checked other adapters and looks like they could be refactored too

kares:
thanks, fine by me - we shall do smt about the new CI failures before merging.
since that isn't caused by this change (and it might seem so).

rsov:
I will try to clean up the CI failures. Once the [ i18n issue] (https://github.com/svenfuchs/i18n/issues/447) with  `symbilize_keys` gets fixed, as most of the CI failures are caused by it

kares:
@rsov was about to merge, just confirm you did not want this to target 51-stable instead?

> In Active Record 5.1 and up secondary parameter column has been deprecated and dropped.

... if you target 51-stable it will merge over to 52-stable (and master) as well

rsov:
That would make sense to merge it into 51-stable. Should I cherrypick the commits into 51-stable branch and push the changes?

kares:
@rsov I attempted to change the base ... but you will need to (cp and) force-push this based off 51-stable

rsov:
Started new pull request with proper changes

PR #968, opened 2018-12-31
[AR52] Ability to handle nil last_inserted_id result

rsov:
When inserting into view's or tables with no ID, exception will be thrown even though the insert goes fine

Code will return `nil` if the result is blank

### DIFF
--- a/lib/arjdbc/db2/adapter.rb
+++ b/lib/arjdbc/db2/adapter.rb
@@ -753,11 +753,10 @@ module ArJdbc
 
     # alias_method :execute_and_auto_confirm, :execute
 
-    # Returns the value of an identity column of the last *INSERT* statement
-    # made over this connection.
+    # Returns the value of an identity column of the last *INSERT* statement made over this connection.
     # @note Check the *IDENTITY_VAL_LOCAL* function for documentation.
-    # @return [Fixnum]
-    def last_insert_id
+    # @return [Integer, NilClass]
+    def last_inserted_id(result)
       @connection.identity_val_local
     end
 


lib/arjdbc/jdbc/adapter.rb
```diff
@@ -358,6 +358,10 @@ def update_lob_value(record, column, value)
       # @return [Integer, NilClass]
       def last_inserted_id(result)
         if result.is_a?(Hash) || result.is_a?(ActiveRecord::Result)
+          # If table does not have primary key defined
+          return nil if result.first.blank?
+
```
kares: :scissors: 
did you review if any one the (MRI) adapters do a similar thing?
cause if not we might be approaching this from the wrong end.
also if we can avoid using `blank?` here - since its not clear whether its and empty `result.first` Hash or a `nil`

enebo: @kares @rsov If this is only captures test issues with a single adapter I would also prefer this is not in generic part of the code base.  Tough if you are only examining a single db but unless one of the failing CI tests goes green or we can tell there is missing coverage amongst all the adapters....we should probably try and isolate.  Otherwise in 2-3 years we won't know why this line is there.

rsov: @kares @enebo I can move this into the db2/as400 code. That is probably a much better way of doing this

enebo: @rsov for me it will lower the barrier to acceptance quite a bit.  I may still comment but I am much more open to you making progress you are happy with than polishing an unreleased/unsupported diamond.

OTOH if you do happen to notice a lack of coverage in AR/AR-JDBC and found a general problem then please go through the extra trouble of putting it in the generic place.

dr-itz: Unless I'm missing something, this implementation of `last_inserted_id` is not used by the "big three". SQLite has its own implementation, PG uses the plain AR implementation, adds something on top, MySQL uses plain AR.

@rsov I wonder if it would work for you just deleting that method from `lib/arjdbc/jdbc/adapter.rb` and rely only on the AR provided method?

rsov:
Good news everyone! There was already a method inside the db2 adapter that does it. Method name and parameters just needed to be updated for AR52 support. Nice and clean solution

enebo:
@rsov I think we would be ok if you wanted to add rails tests to accepted failures part of travis.yml...we can then see progress and potentially catch issues before merging.

rsov:
@edsono You want to run rails tests in junction to our tests as a part of a build but allow failures? That is a good idea. I think I should be able to set that up

enebo:
@rsov if you look at travis you will see we have an internal suite and AR tests running.  Having DB2 in there if you are working on it will help see how you are doing

PR #970, opened 2019-01-02
[postgres] support timestamp arrays

dr-itz:


src/java/arjdbc/util/DateTimeUtils.java
```diff
@@ -577,4 +590,102 @@ private static int extractIntValue(final CharSequence str, int beg, int end) {
...
+            NUMBERS[i] = ((i < 10 ? "0" : "") + Integer.toString(i)).toCharArray();
+        }
+    }
+
+    public static String timestampTimeToString(final ThreadContext context,
+                                               final IRubyObject value, DateTimeZone zone, boolean withZone) {
+        if (value instanceof RubyFloat) {
+            final double dv = ((RubyFloat) value).getValue();
+            if (dv == Double.POSITIVE_INFINITY) {
+                return "infinity";
```
kares: what I do not like is that this piece of utility is now postgres specific.
we shall at least mention that as the method's javadoc so it does not get 'unexpectedly' used elsewhere.

src/java/arjdbc/postgresql/PostgreSQLRubyJdbcConnection.java
```diff
@@ -305,46 +340,9 @@ protected void setTimestampParameter(final ThreadContext context,
...
-            timeValue = timeInDefaultTimeZone(context, timeValue);
-            DateTime dateTime = timeValue.getDateTime();
-            timestamp = new Timestamp(dateTime.getMillis());
-
-            if (timeValue.getNSec() > 0) timestamp.setNanos((int) (timestamp.getNanos() + timeValue.getNSec()));
-
-            statement.setTimestamp(index, timestamp, getCalendar(dateTime.getZone()));
-        }
-        else {
-            setTimestampBC(statement, index, timeValue);
```
kares: this BC thing has been tricky back and forth. you feel its all good, no regressions in terms of passing tests?

src/java/arjdbc/util/DateTimeUtils.java
enebo: @dr-itz yeah I agree with @kares.  Maybe even move this into postgresql packaged class as static utility methods.  We have been trying to make sure any special logic does not get confused and used by another adapter.  Even if it works then at a later point we won't know if any logic is specific to one thing or all things.

dr-itz: agreed, the infinity stuff doesn't belong here; will move to PG specific code

src/java/arjdbc/postgresql/PostgreSQLRubyJdbcConnection.java
dr-itz: yes, tests are passing. In fact I did the normal timestamp handling before the array thing just to make sure.

dr-itz:
ok, new version pushed out, moving the infinity stuff to a PG specific helper class and adding some minimal javadoc.

Note: Travis is completely broken...guess it tries to install bundler 2.0 😞

dr-itz:
For travis fix, see #972 

A build with that commit on top for this PR: https://travis-ci.org/dr-itz/activerecord-jdbc-adapter/builds/474982511

dr-itz:
Crap...while the Rails tests run fine (it's green for PG for the first time actually), this breaks the ARJDBC local tests. Will look into it

enebo:
@dr-itz those local tests are quite a bit older than AR tests.  Definitely make sure it is not testing no longer supported AR behavior.

dr-itz:
@enebo Fixed with #973

enebo:
@dr-itz yeah merged...we can watch the CI results pour in!

src/java/arjdbc/postgresql/PostgreSQLRubyJdbcConnection.java
dr-itz: @kares yes, see below. It's fixed with https://github.com/jruby/activerecord-jdbc-adapter/pull/973 . the builds after that look ok

dr-itz: no worries. yeah...see #974...that should get 52-stable green...my plan is to iron out a few other things and then start with Rails 6 tests

PR #978, opened 2019-01-12
[postgres] fix bind params length

### LINKED ISSUE #977: Queries with bind param size between 32767 - 65535 are improperly using prepared statements with Rails 5.2.2
## Versions Used

- Rails 5.2.2
- AR-JDBC 52.0
- JRuby 9.1.17.0
- PostgreSQL 9.3

## Problem

[This change](https://github.com/rails/rails/commit/213796fb4936dce1da2f0c097a054e1af5c25c2c) to ActiveRecord caused all prepared statements to use bind params for IN clauses. This doesn't work for large queries, and was followed up with [this change](https://github.com/rails/rails/pull/33844) and [this change](https://github.com/rails/rails/pull/33878) to fall back to unprepared statements when the "bind params limit" is exceeded.

Unfortunately the value chosen for the bind params limit was 65535, which doesn't work for JDBC drivers. I know very little about this, but some digging brought me to [this article](https://makk.es/blog/postgresql-parameter-limitation/), which suggests that the PostgreSQL JDBC driver's limit is 32767. At the moment any queries with 32767 > bind params length < 65535 are using prepared statements when they shouldn't be in Rails 5.2.2.

## Evidence

I don't have a link to a reproduction, but running the following queries with any model (here I've used "Thing") in a Rails console running the Rails 5.2.2 and AR-JDBC 52.0 should show you what I mean:

`Thing.where(id: [1] * 30000)` succeeds
`Thing.where(id: [1] * 65000)` succeeds
`Thing.where(id: [1] * 45000)` fails with PSQLException "This connection has been closed"

## Solution

I _think_ the solution is to override the `bind_params_length` method in `connection_adapters/abstract/database_limits.rb` shown in [this commit](https://github.com/rails/rails/pull/33878/commits/b571c4f3f2811b5d3dc8b005707cf8c353abdf03). For AR-JDBC I think it should be 32767 instead of 65535. I'm not familiar enough with Github to give it a try on my own, although if no one else can get to it soon I'll do my best! 

Thanks!

dr-itz:
pgjdbc doesn't like more than 32767 bind params.

Fixes #977

### DIFF
--- a/test/bind_params_length_test.rb
+++ b/test/bind_params_length_test.rb
@@ -1,0 +1,17 @@
+require 'test_helper'
+require 'models/topic'
+
+class BindParamsLengthTest < Test::Unit::TestCase
+  def setup
+    TopicMigration.up
+    Topic.create!(title: 'blub')
+  end
+
+  def teardown
+    TopicMigration.down
+  end
+
+  def test_bind_param_length
+    assert Topic.where(id: [1] * 75000).first
+  end
+end
--- a/lib/arjdbc/postgresql/adapter.rb
+++ b/lib/arjdbc/postgresql/adapter.rb
@@ -690,6 +690,10 @@ module ArJdbc
       @local_tz ||= execute('SHOW TIME ZONE', 'SCHEMA').first["TimeZone"]
     end
 
+    def bind_params_length
+      32767
+    end
+
   end
 end
 
--- a/test/db/mysql/bind_params_length_test.rb
+++ b/test/db/mysql/bind_params_length_test.rb
@@ -1,0 +1,2 @@
+require 'db/mysql'
+require 'bind_params_length_test'
--- a/test/db/postgresql/bind_params_length_test.rb
+++ b/test/db/postgresql/bind_params_length_test.rb
@@ -1,0 +1,2 @@
+require 'db/postgres'
+require 'bind_params_length_test'


enebo:
@dr-itz This also applies in 5.0 too doesn't it?

dr-itz:
@enebo surprisingly no, bind params for "IN" were done in AR 5.2 along with `bind_params_length`

test/bind_params_length_test.rb
```diff
@@ -0,0 +1,17 @@
...
+    TopicMigration.up
+    Topic.create!(title: 'blub')
+  end
+
+  def teardown
+    TopicMigration.down
+  end
+
+  def test_bind_param_length
+    assert Topic.where(id: [1] * 75000).first
```
jessicaspeir: One thing to note is that 75000 parameters should work in AR 5.2 before this change, it only fails for parameter sizes between 32767 and 65535.

dr-itz: @jessicaspeir Oops...you're right of course...forgot to stage (once  again). I used 75000 initially to have the test passing before the change to verify the test itself was basically ok.

@enebo Should I do another PR or may I ask you to change the value to 35000? (I just replaced that glued in battery in my MacBook and then had the silly idea to update to Mojave...and not everything is working as it should just yet... 😉 )

enebo: @dr-itz yeah I can change it!  Will do shortly.

enebo: @jessicaspeir Thanks for catching this.  We appreciate another set of eyes...

enebo: Changed in fc44fedf

jessicaspeir: No problem, thanks for fixing it so quickly!

PR #979, opened 2019-01-12
[test] exclude a time precision test for intermittent failures

dr-itz:
Same as with DateTime precision...


kares:
what JRuby version did you check? just asking whether its similar to this (works in 9.1.17): https://github.com/jruby/jruby/issues/5556

dr-itz:
@kares Got you a reproducer there, but it's kinda crazy...

Issue #980, opened 2019-01-22
Rails 6 support


dr-itz:
This is to track/coordinate support for Rails 6.

/cc @enebo @kares @rdubya

dr-itz:
So..from the remaining few test failures:

* `CoreTest#test_pretty_print_persisted`: turns out this shows up on 5x-stable too when running on JRuby 9.2.x, 9.1.x doesn't show the problem. I'll try and see if I can extract a test
* `LogSubscriberTest#test_vebose_query_logs`: cannot reproduce locally just yet...will look closer in the next few days or so
* There's a query cache problem for PG I'll look at
* Sometimes we get this, might be jruby/jruby#5519

```
Exception in thread "Ruby-0-Thread-92: /home/travis/.rvm/gems/jruby-9.2.5.0/bundler/gems/rails-da44a9c0490a/activerecord/test/cases/reaper_test.rb:73" java.lang.NullPointerException
	at org.jruby.internal.runtime.ThreadService.getMainThread(ThreadService.java:232)
	at org.jruby.RubyThread.exceptionRaised(RubyThread.java:1762)
	at org.jruby.internal.runtime.RubyRunnable.run(RubyRunnable.java:112)
	at java.lang.Thread.run(Thread.java:748)
```

dr-itz:
Quick update: the failing pretty print test is about a Time being off. I can make the test pass by setting `TZ=-1:00`. No idea yet if it's fixture parsing or printing. That's for tomorrow, unless someone beats me to it (hint, hint, I love date/time handling sooo much)

* Expected: `written_on: 2003-07-16 14:28:11 UTC`
* Actual: `written_on: 2003-07-16 13:28:11 UTC`
* Fixture: `written_on: 2003-07-16t15:28:11.2233+01:00`

dr-itz:
The pretty print problem should be fixed by jruby/jruby#5583 if that was actually the only problem. Let's see how that goes.

enebo:
@dr-itz I just restarted last arjdbc travis job after that merge which should pull in latest and greatest....

dr-itz:
@enebo it's pinned to 9.2.5.0....the jruby-head are currently disabled. To test it out, I pushed to my branch again with two jruby-head builds: https://travis-ci.org/dr-itz/activerecord-jdbc-adapter/builds/486663038
[...]

PR #981, opened 2019-01-22
First batch of Rails 6 changes

dr-itz:
See #980


dr-itz:
Basic stuff works and our local test suite passes, the Rails test are a horrible mess. For now I changed Travis not to run most stuff:
* Java 11 is currently just wasting Travis time, let's fix it on Java 8 first
* Same goes for jruby-head
* Rails tests still give way too many errors, but that probably requires just a few fixes to get errors down to a reasonable level

enebo:
@dr-itz yeah seems reasonable.  I also am somewhat surprised we will not be getting hit by i18n using refinements by this but we nearly have that resolved.

enebo:
@dr-itz ah great.  We still plan on making the refinements version work very soon and we ran into another gem which also had swapped to refinements (don't remember which though...).  Good to have one less short term thing holding us from running Rails without special instructions.

PR #983, opened 2019-01-26
Third batch of Rails 6 changes

dr-itz:
This one is all about SQLite3

#980


enebo:
@dr-itz one thing we want to do as a project is review how having n supported branches is working from a support standpoint.  Your progress on 6 is pretty amazing so I think it seems to be working but we should review how onerous this was versus the all-in-one way.  For me the only downside is lots of merge commits.  Those are trivial to do though so I have been overall happier.

enebo:
@dr-itz I will the same way.  On master I intended on following 6.0 development much earlier than when we actually started but I never changed any of the testing runs to track it.  With that said, I tend to agree so long as we are not able to be part of Rails HEAD development process switching master later would be a better approach.

PR #986, opened 2019-02-06
Random Rails 6 bits

dr-itz:
Once JRuby 9.2.6.0 is released and the .travis.yml is changed to use it, this _should_ turn CI green for Rails 6.0.0.beta1.


kares:
yay, would be great to get 9.2.6 out!

PR #987, opened 2019-02-11
[ci] use JRuby 9.2.6.0, use Travis provided bundler

dr-itz:


dr-itz:
fwiw the failing PG tests actually pass sometimes:
https://travis-ci.org/dr-itz/activerecord-jdbc-adapter/builds/491862840

I'v seen it fail locally while running the whole suite, but only once. 40 tries of just that test: works

Issue #831, opened 2017-11-13
jndi tests fail completely in 50.0


enebo:
In commit 94849a9b I removed the travis testing of jndi: https://travis-ci.org/jruby/activerecord-jdbc-adapter/jobs/301579152.

I don't actually know which db this uses to test against but I think using jndi pooling etc is something we will want to support again.

PR #992, opened 2019-03-27
MySQL JNDI fixes

dr-itz:
See #831


test/jndi_mysql_test.rb
```diff
@@ -24,26 +24,22 @@ def setup
...
-      assert_false pool.active_connection? if pool.respond_to?(:active_connection?)
+      assert_nil pool.active_connection? if pool.respond_to?(:active_connection?)
 
-      adapter_class.any_instance.expects(:configure_connection).once
+      # called twice since there's a version check during initialize that needs a connection
+      adapter_class.any_instance.expects(:configure_connection).at_least_once
       ActiveRecord::Base.connection.exec_query "SELECT VERSION()"
     end
 
-    test "configures on demand (since it's lazy)" do
```
kares: this won't pass? 
not sure how much it matters but the problem with 'older' AR was that a conn was being checkout from the middleware regardless of whether a conn was about to be used during a request ... thus with AR callbacks we worked around such situations by only checking out on first actual use.

src/java/arjdbc/jdbc/ConnectionFactory.java
```diff
@@ -49,84 +43,3 @@
     Connection newConnection() throws SQLException;
 
 }
-
-class DataSourceConnectionFactoryImpl implements ConnectionFactory {
```
kares: don't recall but these are no longer used than?
... do we still support `data_source:` option along side with `jndi:`

src/java/arjdbc/jdbc/RubyConnectionFactory.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+import org.jruby.RubyObject;
+import org.jruby.RubyString;
+import org.jruby.runtime.ThreadContext;
+import org.jruby.runtime.builtin.IRubyObject;
+
+// @legacy ActiveRecord::ConnectionAdapters::JdbcDriver
```
kares: was really legacy ... think that I might have been hoping to drop it as/if possible.
but honestly I am not sure how much its relevant, pbly worth keeping if needed to JdbcAdapter.

test/jndi_mysql_test.rb
```diff
@@ -24,26 +24,22 @@ def setup
...
 
-    test "configures once" do
+    test "configures connection" do
       pool = ActiveRecord::Base.connection_pool
-      assert_false pool.active_connection? if pool.respond_to?(:active_connection?)
+      assert_nil pool.active_connection? if pool.respond_to?(:active_connection?)
 
-      adapter_class.any_instance.expects(:configure_connection).once
+      # called twice since there's a version check during initialize that needs a connection
+      adapter_class.any_instance.expects(:configure_connection).at_least_once
```
kares: this is a bit gross ... and smells but we should have a special configuration to skip configure_connection for good.

kares:
+1 happy too - very appreciated work!
although some tests indicates that we might be in a semi-supported state (kind of working but not ideal). 
still okay by me - have always spent too much time digging into details when looking into JNDI support.

test/jndi_mysql_test.rb
dr-itz: nope, won't pass...problem is, somewhere in the 5.0 line Rails added a version check in initialize() of the abstract mysql adapter. This version check does nothing more than connect to the db, select the version and complain that it is too old...oh well...for fun I just removed it the Rails code, to have this passing...

dunno...one thing we could do is to set version to a dummy value high enough before calling super, reset it to nil afterwards so the next real call that needs a version does actually load it. I tried that too and it seems to work, but it's a bit dirty. Just noticed that in Rails 6 they moved it to a `check_version` function that can be overriden with nothing...exactly to have it lazy again..

so...I like neither of the options too much, but being really lazy would be good.

test/jndi_mysql_test.rb
dr-itz: agreed it's ugly...I think I try the version check dummy thing I mentioned above...

src/java/arjdbc/jdbc/ConnectionFactory.java
dr-itz: yeah..that one I tripped over...there's a class in its own file, DataSourceConnectionFactory that is actually used, this one is stale. So I moved the other two out to separate files and deleted that one.

src/java/arjdbc/jdbc/RubyConnectionFactory.java
dr-itz: no idea...I decided not to look 😆

dr-itz:
ok, new version that should be lazy...I like it better now

test/jndi_mysql_test.rb
dr-itz: passes now

test/jndi_mysql_test.rb
dr-itz: untouched in the new version

PR #993, opened 2019-03-30
[test] fix some ArgumentError: argument out of range

dr-itz:
This is already in 52-stable. With latest Rails 5.0.x, it happens too.


kares:
yep, did notice and got a bit confused. thanks

dr-itz:
btw. I've seen 52-stable is a bit too red for my taste...here, most of the stuff can be fixed by cherry-picking some commits from master....I'll prepare series with backports for 50, 51 and 52 in the next few days

Issue #965, opened 2018-12-11
Java::JavaLang::IllegalAccessError (tried to access method org.postgresql.jdbc.PgResultSetMetaData.getField(I)Lorg/postgresql/core/Field; from class org.postgresql.jdbc.PgResultSetMetaDataWrapper)


mengu:
hello everyone

i've been trying to use tomcat's connection pooling feature and i get the error in the title. 

**rails version:** 5.2.0
**jruby version:** 9.2.5.0 (and 9.2.0.0)
**ar-jdbc-adapter version:** master
**tomcat version:** 9.0.13 (also tried with glassfish 5 - build 25)
**postgresql jar version:** 42.1.4 (also tried with 42.2.5)

here are the steps i've followed:

- changed config/database.yml to the following:
```
production:
  adapter: postgresql
  jndi: jdbc/myJndiSource
```

(i have also tried setting jndi to `java:comp/env/jbc/myJndiSource`)

- configured my web.xml and tomcat's context.xml accordingly.
- added postgresql jar file to lib dirs of tomcat and glassfish. 

result is the following stack trace:
```
  Java::JavaLang::IllegalAccessError (tried to access method org.postgresql.jdbc.PgResultSetMetaData.getField(I)Lorg/postgresql/core/Field; from class org.postgresql.jdbc.PgResultSetMetaDataWrapper):
org.postgresql.jdbc.PgResultSetMetaDataWrapper.getField(PgResultSetMetaDataWrapper.java:21)
arjdbc.postgresql.PostgreSQLResult.columnTypeMap(PostgreSQLResult.java:81)
arjdbc.jdbc.JdbcResult.toARResult(JdbcResult.java:128)
arjdbc.postgresql.PostgreSQLResult.toARResult(PostgreSQLResult.java:171)
arjdbc.postgresql.PostgreSQLRubyJdbcConnection.mapQueryResult(PostgreSQLRubyJdbcConnection.java:282)
arjdbc.jdbc.RubyJdbcConnection$15.call(RubyJdbcConnection.java:1100)
arjdbc.jdbc.RubyJdbcConnection$15.call(RubyJdbcConnection.java:1091)
arjdbc.jdbc.RubyJdbcConnection.withConnection(RubyJdbcConnection.java:3438)
arjdbc.jdbc.RubyJdbcConnection.withConnection(RubyJdbcConnection.java:3404)
[...]

PR #994, opened 2019-03-31
RFC: remove the PG result set wrapper

dr-itz:
* AR uses the OID to map to named types for which handlers are registered
* JDBC can return the named types with getColumnTypeName()
* For normal (non-array, non-rage, etc.) this already works
* by adding handler based on the name as well as the OID,
  getColumnTypeName() can be used for everything. OID based mapping
  is still required for SchemaStatements.
* loading additional types needs to be adapted to handle OIDs as well
  as type names.
* pgjdbc returns two additional type: 'serial' and 'bigserial'. They're
  just int4 and int8 with auto-increment on top
* only trick thing is "fmod". luckily this is only used for "numeric"
  and can be re-constructed by using getPrecision() and getScale()
* one test needs to be excluded because it checks for a message containing
  an OID (number) instead of name (string) we get now

See #965 

**NOTE**: this is against master because master is otherwise green. Would make sense to rebase against 50-stable. If it's any good that is :)


rdubya:
I haven't looked at this stuff in a long time, but I thought the reason we needed the specific oid was because the type of data inside the array was linked to that. Does the name include that info?

enebo:
@dr-itz looks like this creates one extra query for one test?  Personally, if @rdubya's concern is not an issue it would be great to remove the reflective hack since the jdbc adapter guys are not keen on exposing oid.

dr-itz:
@rdubya Arrays are handled fine. They needed a bit of special code to register by their names as well. OID is unique, typname within typnamespace too. That reminds me that we need to look at the namespace too to be really correct (note that AR misses it too). I'll check..

@enebo No additional queries that I can see (well, except those done internally by pgjdbc, but they don't show). The only thing as far as testing goes is the exclude I added...and that is just because the message "...regclass..." doesn't match "...\d+..." 😉

enebo:
@dr-itz I have not checked recently but I thought master was green and we appear to have on fail with this PR (extra query).

dr-itz:
@enebo ah..that one is MySQL, so it's unrelated...it'll probably pass if you restart the job

enebo:
@dr-itz can you reapply to 50-stable?  Looks like @rdubya question has been answered and your other PR was merged so things should be pretty much green on 50 now.

dr-itz:
@enebo will do, but thinking about it a bit more, the namespace thing needs to be fixed before this is ready. It could possibly select the wrong type based on the name alone...the lookup needs to handle three cases:
* by OID for SchemaStatements. Check
* by name w/o namespace. Half-check; needs to make sure that only the right namespaces are searched in. Nothing big, just an additional WHERE
* by name w/ namespace (pgjdbc can return this). ToDo; split by ".", possibly unquote than have a query that joins the `pg_namespace` table. Seems easy enough

enebo:
@dr-itz I am hoping to revv all 50+ releases this week.  So this can possibly go in for next point release unless you happen to solve this in next day or two.

dr-itz:
@enebo I'm having some more time to look into it Thursday/Friday...currently I'm preparing some minor fixes for 51-stable and 52-stable...first PR soon(ish)

dr-itz:
Ok, namespace handling implemented and rebased onto 50-stable. No regressions so far...

Note on merges: later versions of Rails changed the signature of `load_additional_types()` by dropping the first argument...that's all

dr-itz:
https://travis-ci.org/dr-itz/activerecord-jdbc-adapter/builds/516662643

mengu:
thank you very much for this @dr-itz.

PR #998, opened 2019-04-02
51.x fixes

dr-itz:
One fallout from the 50-stable merge, CI Rails version update, excludes update and one for SQLite3 that probably doesn't trigger for 51-stable, but it sure does for 52-stable.

The remaining two errors on PG I have no idea about yet...test and rake tasks are the same as in Rails 5.2.

This will likely result in conflicts in the 52-merge. Anyways, once this is in, another PR for 52-stable will follow soon after.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,9 +92,9 @@ matrix:
         mariadb: '10.1'
       env: DB=mariadb PREPARED_STATEMENTS=true
     # Rails test-suite :
-    - env: DB=mysql2     TEST_PREFIX="rails:" AR_VERSION="v5.1.6.1" # PS off by default
-    - env: DB=mysql2     TEST_PREFIX="rails:" AR_VERSION="v5.1.6.1" PREPARED_STATEMENTS=true
-    - env: DB=mysql2     TEST_PREFIX="rails:" AR_VERSION="v5.1.6.1" DRIVER=MariaDB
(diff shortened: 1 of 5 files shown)


enebo:
@dr-itz I merged across along with the mysql jdbc adapter update.  We shall see how well that went...

dr-itz:
@enebo 👍 I'll start preparing a few fixes for 52-stable then

enebo:
@dr-itz thanks.  Right now I am looking into the jruby-head quote Array issue.  I noticed this in some work I am trying to accomplish so I it looks pretty common for Rails apps.  Hopefully you will see those turn green today (or it is still up for grabs but I am fairly certain it is a JRuby bug on master).

dr-itz:
@enebo just a heads-up...with plain 52-stable: when running rails:test_postgresql (haven't tried the others) I'm seeing extreme slowdowns, some "E"s during test, but they never report since I get an OOM. Problems go away when using -J-Xmx1024M. That has never been an issue before...so we're leaking...

kares:
@dr-itz is this after the JDBC driver update?

dr-itz:
@kares nope...I don't know too much about it yet...just that it's mostly char[], ByteList, HashEntry...writing a heap dump on OOM didn't work, just hangs the VM...it _might_ be the Rails version..will try to narrow it down this evening...

enebo:
@kares @dr-itz postgresql I did not prop the pg jdbc driver update past 50-stable yet so it shouldn't be that.

PR #1000, opened 2019-04-05
52.x fixes

dr-itz:


enebo:
@dr-itz I fixed our issue in JRuby so jruby-head now should pass everywhere.  Did you happen to still see the issue of memory+slowness you saw before?

dr-itz:
@enebo 👍 on the jruby-head fixes.

yeah, the memory+slowness is still there, but...I tried going back to 52.0 with Rails 5.2.0...same problem. Raising the memory limit really fixes everything and `rails:test_postgresql` finishes in 4-5min. I _guess_ memory was always close to the limit. The one thing that actually changed in my env are the gems used for the tests. When Rails updated some of their dependencies and bundler complained, I just deleted my local Gemfile.lock...

Still, there are some leaks, but that might be Rails. VisualVM shows around 180 live threads when tests are close to finished. Most of it is the ConnectionPool Reaper from Rails
52-stable-523

dr-itz:
@enebo The ConnectionPool::Reaper threads are a leak in Rails. I'm working on a patch for Rails now...For PG, I'm currently down to 80 or so. There's also some ConnectionPools that are leaked completely...

enebo:
@dr-itz ah well you found the source so that is good news.  We appreciate the time you have been spending.  It has taken a lot of work off our plates this spring.

dr-itz:
So here is what I've come up with: https://github.com/dr-itz/rails/tree/fix/stop-reaper-52
It's for Rails 5.2.x, but I'll port to master in the next few days and try to get it merged

enebo:
@dr-itz It will be great if they accept it.  My only question is whether that double break check is needed or whether it just alleviates waiting an extra sleep period.

dr-itz:
Here we go: https://github.com/rails/rails/pull/35947

PR #1001, opened 2019-04-06
[perf] set a fetch size to reduce peak memory usage

dr-itz:


enebo:
@dr-itz I am a little confused by this enhancement.  So this change will limit the default to only fetch at most 100 but we can enable it via config to be any value?  Should 100 be a default size here?  I think even if this is a good default we need an option to explicitly disable the fetch size (e.g. -1 will not set fetch limit).

dr-itz:
Note that it's not a limit, just tells the driver how many rows to fetch at once from the server. For most drivers, the default is to fetch all rows at once which can use quite a bit of memory. 

Anyway, it's just a hint for the driver. 0 means default behavior (fetch all at once), so no special case like -1 is necessary. We could also not set it when it's 0...and maybe even leave the default at 0...dunno. But the config option is important for performance tuning.

A fetch size of 100 is a good default, mostly from experience, also it won't affect most queries in a Rails app. And for anything that fetches a few thousand rows it will decrease memory usage _and_ increase speed. The "perfect" fetch size of course depends on the data being fetched, but the difference between 100 and e.g. 500 is usually so small in terms of performance that 100 or 200 is good enough.

enebo:
@dr-itz thanks for the explanation.  I personally feel like the default should be the JDBC default but I guess I am not someone who has a lot to bring to the discussion.  @kares @rdubya what do you guys think about this?

kares:
no strong opinion here - would maybe leave the default (explicit) 'fetch-less' and allow tuning with a config option (+ driver specific properties). been a while so I believe we should review jdbc driver (mysql/pg) docs on this, or maybe even how the native ones behave. also having some raw benchmark numbers on queries where this makes sense, would be useful?

rdubya:
I don't know enough about how that will affect things to have too much of an opinion. One question that does pop up in my mind though is how this works with things like pg_badger. We can't use prepared statements because of pg_badger since the connections to the actual database are shared among the connections to pg_badger, so the statement might get prepared on one connection and then executed on another. Will this fetching setting have the same problem or does it count as one continuous connection?

enebo:
@dr-itz @kares @rdubya  I guess based on the collective feedback we have 2 leave default unless we learn something, 1 no opinion but wonders something (pg_badger), and your submission which would be to enable this.  Seems best path for the moment would be to accept this PR BUT change the default to 0 (which will be default fetch behavior).  This would allow this to be set and if we learn that we have good bench results or knowledge that a particular fetch size is best then a second PR can set that value.

dr-itz:
Thanks for the feedback. pg_badger and friends is something I haven't considered. I changed it now to a complete opt-in behavior, not event trying to set the fetch size if it's 0 as there might be driver bugs around.

enebo:
@dr-itz thanks again.  At some point we will need to update our docs on the project and provide current info for configuration: https://github.com/jruby/activerecord-jdbc-adapter/wiki/Configuration.

rdubya:
Just wanted to correct something I said, the technology we use is called pgbouncer not pg_badger 🤦‍♂️, but the comment still applies. pg_badger is another tool that wouldn't be involved in this case.

PR #1004, opened 2019-04-12
Make `check_version` public

deivid-rodriguez:
Since Rails now needs it public. Since https://github.com/rails/rails/pull/35795.


enebo:
Thanks for the PR.  I did not read that issue closely but I am guessing beta4/final will have something which will pass a test by making this change.

PR #1008, opened 2019-04-16
Work towards using activerecord-sqlserver-gem for mssql support

rdubya:
This is still very much a WIP. I pushed it up so that other people could get eyes on it. My approach has been to take a similar route that we took with the "big 3" adapters and try to reuse the gems that have support built out already and just override the pieces that don't work with jdbc with the hopes that we can get changes into future versions of the gems to clean up all the overrides/hacks.

This PR needs https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/694 to make it very far through the tests as well as a bunch of other supporting changes that I haven't cleaned up enough to push up yet.

I currently have my repos hacked up so that I can run the activerecord-sqlserver-adapter gem's test suite against this code (which is a combination of the active record test suite and their own tests) and my last run had `4917 runs, 14098 assertions, 11 failures, 8 errors, 14 skips`, most of the failures left are around transaction isolation and date/time precision. There are also some tests that need to be excluded because of mri/jruby differences and driver differences.

One thing that probably won't work on initial release is supporting UUID columns as auto-generated values. I think it can be done, but the way the values are generated requires some workarounds to return them.

If anybody is interested in how I set my stuff up to be able to run these tests I can try to get it documented/cleaned up enough that the changes can be committed.


rdubya:
@kares @enebo I think this is basically good for the mssql stuff if you want to look at it, I need a new release of 50.x before I can get the activerecord-jdbcsqlserver-adapter gem to a point that it can be released

rdubya:
@kares @enebo The test runs actually look better than the last runs against 50-stable so I think test-wise we are ok with this if you think the changes are ok.

lib/active_record/connection_adapters/mssql_adapter.rb
```diff
@@ -1 +0,0 @@
-require 'arjdbc/mssql'
```
kares: do we not want the `adapter: mssql` magic to work? (and require sqlserver adapter parts)

lib/active_record/connection_adapters/sqlserver_adapter.rb
```diff
@@ -1 +0,0 @@
-require 'arjdbc/mssql'
```
kares: if sqlserver adapter has this file we should keep it as a potential alternative working with `adapter: sqlserver`

rakelib/02-test.rake
```diff
@@ -58,7 +58,6 @@ end
 test_task_for :Derby, :desc => 'Run tests against (embedded) DerbyDB'
 test_task_for :H2, :desc => 'Run tests against H2 database engine'
 test_task_for :HSQLDB, :desc => 'Run tests against HyperSQL (Java) database'
-test_task_for :MSSQL, :driver => :jtds, :database_name => 'MS-SQL (SQLServer)'
```
kares: how do we run tests than?

rdubya: The tests will all run from the sqlserver codebase. We can try to set something up to run from this side but since the dependency goes from the sqlserver gem to this one, its kind of hard to have the tests run on this side. Ideally the java pieces would have all been in the sqlserver codebase but that would be another project to be able to get things compiled then.

lib/active_record/connection_adapters/sqlserver_adapter.rb
rdubya: Yeah, pretty much the only reason there needs to be changes to this repo is because of the Java source needing to live in here for compiling.

kares: okay, would hope if the fork is to be maintained that the .java source would move there.
fine for now but this is the way AR-JDBC exts used to work in the past and I do not see a reason why keep it here (except if the Java base class API would not be 'stable'). before doing that its okay to have the native part in as I see a few minor pieces we should maybe extract away (as it gets merged - hopefully before release).

rdubya: I don't remember exactly but I think I had chatted with @enebo about it and its not that easy to separate out the java classes to a separate compile step. I suppose I could look into doing that but at this point I just want to get this out so that we can officially support rails 5.0 and I can get our (Finalsite's) app upgraded. I hope 5.1 and 5.2 will require minimal effort so I'd like to get into those once we get this one into the wild and get other people testing it out.

rdubya:
Is there anything else that I need to do with this to get it merged?

PR #1010, opened 2019-04-24
Fixes for 6.0.0.rc1, cleanups, refactorings

dr-itz:
Note: all commits starting with "modernize:" were done automatically by IntelliJ

### DIFF
--- a/src/java/arjdbc/sqlite3/SQLite3RubyJdbcConnection.java
+++ b/src/java/arjdbc/sqlite3/SQLite3RubyJdbcConnection.java
@@ -225,7 +225,7 @@ public class SQLite3RubyJdbcConnection extends RubyJdbcConnection {
                         IRubyObject[] args = new IRubyObject[] {
                             cachedString(context, indexTableName), // table_name
                             cachedString(context, indexName), // index_name
-                            nonUnique ? runtime.getFalse() : runtime.getTrue(), // unique
+                            nonUnique ? context.fals : context.tru, // unique
                             currentColumns = RubyArray.newArray(runtime, 4) // [] column names
                         };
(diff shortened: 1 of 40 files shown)


src/java/arjdbc/sqlite3/SQLite3RubyJdbcConnection.java
```diff
@@ -516,7 +513,7 @@ protected void setTimestampParameter(final ThreadContext context,
         final int index, IRubyObject value,
         final IRubyObject attribute, final int type) throws SQLException {
 
-        if (value instanceof RubyTime) value = ((RubyTime) value).strftime(TIMESTAMP_FORMAT);
+        if (value instanceof RubyTime) value = ((RubyTime) value).strftime(context, TIMESTAMP_FORMAT);
```
kares: did you check since what JRuby version the method exists?

kares:
LGTM, don't like the getter inlines that much.

src/java/arjdbc/sqlite3/SQLite3RubyJdbcConnection.java
dr-itz: since 9.2.0.0 so we're good

dr-itz:
yeah...looking a bit closer at that commit, it's a bit ugly in places, assigning to a local var first...should have manually fixed that after the IntelliJ magic...can still do. Or drop the commit. It was one of those performance suggestions...

enebo:
@kares @rdubya @dr-itz I am rolling out 6.0.0.rc1 probably tomorrow or next day before Railsconf starts.  Speak now or forever speak your piece.

PR #1015, opened 2019-07-09
Build against AR 6-0-stable instead of master

Confusion:
Fixes build failures due to AR moving on and those changes not being handled yet

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ jdk:
   - openjdk8
 env:
   global:
-    - AR_VERSION="master"
+    - AR_VERSION="6-0-stable"
 matrix:
   allow_failures:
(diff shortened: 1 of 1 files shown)


Confusion:
Most builds fail because mpapis's rvm key has expired

Confusion:
I don't see where I could make a change so this is done, but the RVM website says to import these two keys:
`gpg2 --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB`

dr-itz:
👍 for the changes here. The RVM problems seem intermittent. But there are still some problems due to some backports, so I just did #1017 to fix some of them

PR #1016, opened 2019-07-09
Remove unused simplified_type method

Confusion:
Method hasn't been part of ActiveRecord since commit
https://github.com/rails/rails/commit/0b682e4b05c8f58c77c655650af6638c483ac903
(May 2014)

### DIFF
--- a/lib/arjdbc/informix/adapter.rb
+++ b/lib/arjdbc/informix/adapter.rb
@@ -45,20 +45,6 @@ module ArJdbc
       ::ActiveRecord::ConnectionAdapters::InformixColumn
     end
 
-    module ColumnMethods
-
-      private
-      # TODO: Test all Informix column types.
-      def simplified_type(field_type)
-        if field_type =~ /serial/i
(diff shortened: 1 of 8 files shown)


dr-itz:
👍

lib/arjdbc/informix/adapter.rb
```diff
@@ -45,20 +45,6 @@ def jdbc_column_class
       ::ActiveRecord::ConnectionAdapters::InformixColumn
     end
 
-    module ColumnMethods
```
kares: in theory, this must be getting included somewhere ... so maybe keep the remove or find the place

Confusion: Good point. It's used on line 33, but it seems we can remove that entire function, since it's optional. Shall I indeed remove the reference there and the `self.column_selector` function?

Confusion:
Updated version in https://github.com/jruby/activerecord-jdbc-adapter/pull/1039

PR #1017, opened 2019-07-09
More Rails 6 fixes

dr-itz:


dr-itz:
MySQL is still a bit broken, fixes in a day or two...

lib/arjdbc/mysql/adapter.rb
```diff
@@ -40,6 +40,12 @@ def initialize(connection, logger, connection_parameters, config)
         # configure_connection taken care of at ArJdbc::Abstract::Core
       end
 
+      def self.database_exists?(config)
+        !!ActiveRecord::Base.mysql2_connection(config)
```
kares: not sure about this pattern, if we'd like to support JNDI ... connection might be lazy.
isn't there an existing API we could use or maybe do a conn establish + explicit ping?

dr-itz: Nice catch. Haven't thought about that, just blindly copied AR. And thinking about it, this is also a possible connection leak. Will fix up in the next few days

dr-itz: Fixed. Now the check is forced, regardless of JNDI and the connection is closed afterwards

dr-itz:
The failing MySQL tests are interesting:

```
-"developers/1-20190716215152188459"
+"developers/1-20190716215152000188"
```

looking at the last part: `188459` vs `000188`. Notice the `188`....same for the other failing tests. Looks like the milliseconds part becomes the microseconds somewhere along the way. No idea where yet :(

rdubya:
@dr-itz that may be fixed in my branch for the mssql stuff. The DateTime utils stuff was setting it incorrectly.

enebo:
@rdubya something in the time utils stuff with digits after 6 I believe right?

rdubya:
Yeah, this change here: https://github.com/jruby/activerecord-jdbc-adapter/pull/1008/files#diff-47d54384e77767b736ebbe6f06d39ade

dr-itz:
@rdubya thanks, that fixes it! Good thing I didn't go bug hunting yet 😉 

I guess we should extract that one out separately for 50-stable onwards?

kares:
> I guess we should extract that one out separately for 50-stable onwards?

yep let's have it all the way - we actually have a very similar logic (for millis handling) in core JRuby

dr-itz:
> > I guess we should extract that one out separately for 50-stable onwards?
> 
> yep let's have it all the way - we actually have a very similar logic (for millis handling) in core JRuby

done in #1020

dr-itz:
ok, this is all green now

kares:
fine to merge, not 100% sure about the `database_exists?` but I seems good enough.
still have one question looking at this, doesn't the `database_exists?` thingy make sense in 5.X?
or is it a new API in Rails 6.0 (sorry for my ignorance on the topic) ?

dr-itz:
> still have one question looking at this, doesn't the `database_exists?` thingy make sense in 5.X?
> or is it a new API in Rails 6.0 (sorry for my ignorance on the topic) ?

it's a new API added very recently, after the 6.0.0-rc1 release: https://github.com/rails/rails/commit/ee9036ea11d3657e68d187379c92e35b23c6261b

PR #1022, opened 2019-07-18
[jdbc-sqlite3] update to driver 3.28.0

kares:


kares:
let's see how test roll, @dr-itz generally we would only push and release from master.
but if you feel like we need this in other branches as well, let me know ...

dr-itz:
thanks for releasing 3.27.2.1, that's fine by me...the point was basically to have the drivers that fits the released arjdbc 60 rc - what we have in master

tests here fail miserably because #1017 needs to go in first...and together with 50-stable merges it should get green

PR #1024, opened 2019-07-26
[mysql, postgres] respect driver_name from jdbc gems

kares:
this is mostly a preparation for supporting the new major version of MySQL Connector/J (8.0)
... which changed the driver class name

AR-JDBC (at least currently supported AR versions) should work with both major driver versions
... to ease the transition

and since there's no limiting of the jdbc-mysql gem in our helper gem (activerecord-jdbcmysql-adapter), we should delay releasing the MySQL driver (8.0) till new versions of AR-JDBC incorporating support are released

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,6 @@
 language: ruby
 sudo: false
-branches:
-  only:
-  - master
-  - /.*-stable$/
-  - /^test-.*/
-  - /maintenance|support/
-  - /.*dev$/
+
 bundler_args: --without development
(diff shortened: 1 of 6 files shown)


kares:
unfortunately, too many branches, I ended up pushing this to 50-stable early on while trying to:
https://github.com/jruby/activerecord-jdbc-adapter/commit/c996ee363bb55004a55f680e05517ce6fce032af ... as mysql service stopped being available on CI
btw. there's still an issue with starting mariadb, atm ... travis is just getting hard to manage ;(

PR #1025, opened 2019-07-29
[ci] use PostgreSQL 10 (as default)

kares:
match up Rails 6-0-stable, currently this has one failure in the Rails suite (compared to 9.4)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,9 +60,11 @@ matrix:
     - env: DB=mysql2 DRIVER=MariaDB
 
     - env: DB=postgresql PREPARED_STATEMENTS=false INSERT_RETURNING=false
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -59,14 +59,14 @@ matrix:
     - env: DB=mysql2 PREPARED_STATEMENTS=true
     - env: DB=mysql2 DRIVER=MariaDB
 
-    - addons:
-        postgresql: "9.4"
```
dr-itz: I originally did the addons-postgresql per job instead of global (that one must have slipped in) to save some seconds of setup-time on the non-postgres jobs. It's between 3-4s so it won't matter much

kares:
should be defunct ... in favour of https://github.com/jruby/activerecord-jdbc-adapter/pull/1031

PR #1026, opened 2019-07-29
[jdbc-mysql] upgrade to new major version 8.0

kares:
of Connector/J driver, actually 8.0.17 but only 8.0.11 was ga (the firs 8.0.0 release as a pre/beta)

we will need to avoid a bit, since 50.3/51.3/52.2 do not handle the driver name change automatically.

so we'll be releasing new versions of those, note that 1.3 releases are not supported at this point


scystephen:
[CVE-2018-3258](https://nvd.nist.gov/vuln/detail/CVE-2018-3258)is applicable to the gem due to the use of mysql-connector-java-8.0.11.jar.  This vulnerability is resolved in versions after 8.0.12 so a move to 8.0.17 would resolve the issue.

enebo:
@kares is this safe to land?

kares:
@enebo yes its released already (since the merge).
except some new jdbc property names I did not found any AR-JDBC issues using 8.0

enebo:
@kares sorry I read that merge statement backwards and totally spaced out a huge "MERGED" on top of that :)

PR #1030, opened 2019-08-29
Rails 6 test fixes

dr-itz:


dr-itz:
This should be green again. Time to release 60.0?

enebo:
@dr-itz yeah we should definitely release soon.  We really need some image system where we can do some test app as a final QA before releasing.  The tests passing are great but I get nervous unless I try to run something.  I will try and run some stuff locally just for sanity and get something out.  Thanks for keeping it all up to date.

dr-itz:
@enebo makes a lot of sense. Just for the fun of it and because we have to do it anyway at some point, I quickly migrated one of our Rails 5.2 apps to 6.0 with arjdbc-60.0.rc1. And by quickly I really mean it. That was _the_ most painless Rails upgrade ever 🎉 . Everything seems to work and all test pass (including Capybara/Selenium) in MRI and JRuby. It's a small app that does nothing fancy or dirty, but still 😄

enebo:
@dr-itz cool beans.  Thanks for sharing.  It always makes me happy when people say things just work!  It also is probably saying a lot about Rails maturity too that not much is breaking between versions.

PR #1031, opened 2019-09-09
[postgres]: fixes for Rails 6

dr-itz:


dr-itz:
...and it's red again. Seems unrelated, but I'll have a look later ;)

dr-itz:
ok, it's green now, at least in my build: https://travis-ci.org/dr-itz/activerecord-jdbc-adapter/builds/583356318

@kares I also did the PostgreSQL v10 thingy in CI here. I needed at least v9.5 for Rails tests to pass, so I tested v10 while on it. Your #1025 was red for good reason, fixed in the last commit here. Might actually need to pick that one for the 5x-stable line 😉

kares:
so Rails won't pass tests on 9.X anymore?!

let's just have 10 as default (as in #1025 - just copy the changes I am fine closing the PR in favour of yours) and we can override for AR-JDBC's suite to still test some 9.X

dr-itz:
> so Rails won't pass tests on 9.X anymore?!

it's not that bad. It's just one of the new `insert_all` tests that expects to run on PG version supporting "skipping doubles" and that's 9.5. So I left a `test_postgresql` and a `rails:test_postgresql` with 9.5 (instead of 9.4) in there. All the other combinations with PS and INSERT_RETURNING are on 10. I don't like the minimum to be 9.5 when 9.4 is still in support. Maybe we could just to 10 and 9.4 but do a conditional exclude for the one test?

kares:
sure, if you're able to exclude Rails test conditionally or try to bring this upstream on their repo ...

enebo:
I am confused about one thing.  If these fail for us on 9.4 don't they fail for CRuby+pg on 9.4 also?  Or are we supporting older postgresql than Rails 6?

kares:
we used to support older, not sure how much has changed since the 5X releases - we might no longer do so (which is fine)

dr-itz:
The test in question is now excluded on our side depending on the DB features, just like Rails does. So we're green on 9.4 and on 10

PR #1038, opened 2019-10-06
[postgres] use RubyDate

dr-itz:
Since JRuby 9.2.x is required to run Rails 6, we can use the new
RubyDate class directly.


kares:
Thanks Daniel, I picked the first commit on master: https://github.com/jruby/activerecord-jdbc-adapter/commit/794da861f7fe8d9c19df167ec46d8f7d0ad6ab15

there's more we can do here to avoid the (internal) "hack" of wrapping the DateTime instance.
unfortunately, did not put much thinking into exposing some APIs for AR-JDBC during the date.rb to native rewrite - some of those we [could have been re-using here](https://github.com/jruby/jruby/commit/a8807f2b3e4e0f4c69db24e0e8747404d8bc4a49#diff-39bbca646c1ec079b72e5848192dc0efR442), for now we might be stuck with doing a :scissors:

dr-itz:
Thanks for those date-time changes. Looks nicer! 

I just pushed a new version of my change here, mostly removing the custom PG implementation and pushing it to the core. Smaller and faster and not only for PG.

PR #1039, opened 2019-11-01
Remove unused simplified_type method

Confusion:
Method hasn't been part of ActiveRecord since commit
https://github.com/rails/rails/commit/0b682e4b05c8f58c77c655650af6638c483ac903
(May 2014)


Confusion:
(This is a revision/continuation of https://github.com/jruby/activerecord-jdbc-adapter/pull/1016, which it seems I never updated)

PR #954, opened 2018-10-01
Fix RuntimeError: can't add a new key into hash during iteration issu…

pytong:
AR-JDBC's version used:
52.1

version of Rails / ActiveRecord you're running with:
5.2.1

JRuby version (you might include your Java version as well) - jruby -v
jruby-9.2.0.0

include any (related) back-traces (or Java stack-traces) you've seen in the logs

RuntimeError: can't add a new key into hash during iteration
                             []= at org/jruby/RubyHash.java:1033
                mysql_connection at /Users/ptong/.rvm/gems/jruby-9.2.0.0/gems/activerecord-jdbc-adapter-52.1-java/lib/arjdbc/mysql/connection_methods.rb:8
                  new_connection at /Users/ptong/.rvm/gems/jruby-9.2.0.0/gems/activerecord-5.2.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:809
         checkout_new_connection at /Users/ptong/.rvm/gems/jruby-9.2.0.0/gems/activerecord-5.2.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:853
  try_to_checkout_new_connection at /Users/ptong/.rvm/gems/jruby-9.2.0.0/gems/activerecord-5.2.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:832
              acquire_connection at /Users/ptong/.rvm/gems/jruby-9.2.0.0/gems/activerecord-5.2.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:793
                        checkout at /Users/ptong/.rvm/gems/jruby-9.2.0.0/gems/activerecord-5.2.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:521
                      connection at /Users/ptong/.rvm/gems/jruby-9.2.0.0/gems/activerecord-5.2.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:380
             retrieve_connection at /Users/ptong/.rvm/gems/jruby-9.2.0.0/gems/activerecord-5.2.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:1008
             retrieve_connection at /Users/ptong/.rvm/gems/jruby-9.2.0.0/gems/activerecord-5.2.1/lib/active_record/connection_handling.rb:118
                      connection at /Users/ptong/.rvm/gems/jruby-9.2.0.0/gems/activerecord-5.2.1/lib/active_record/connection_handling.rb:90
               block in evaluate at (irb):1

... a way to reproduce :)
[...]

PR #1040, opened 2019-11-10
Various fixes

dr-itz:
This is mainly some "fallout" from ongoing work for Rails 6.1 (i.e. master) support

### DIFF
--- a/test/db/mysql/unit_test.rb
+++ b/test/db/mysql/unit_test.rb
@@ -96,7 +96,7 @@ class MySQLUnitTest < Test::Unit::TestCase
       connection_handler = connection_handler_stub
 
       config = { :jndi => 'jdbc/TestDS' }
-      connection_handler.expects(:jndi_connection)
+      connection_handler.expects(:jndi_connection).with() { |c| config = c }
       connection_handler.mysql_connection config
 
       # we do not complete username/database etc :
@@ -126,7 +126,7 @@ class MySQLUnitTest < Test::Unit::TestCase
       connection_handler = connection_handler_stub
 
       config = { database: 'MyDB' }
-      connection_handler.expects(:jdbc_connection)
+      connection_handler.expects(:jdbc_connection).with() { |c| config = c }
       ::Jdbc::MySQL.expects(:driver_name).returns('com.mysql.CustomDriver')
       connection_handler.mysql_connection config
       assert_equal 'com.mysql.CustomDriver', config[:driver]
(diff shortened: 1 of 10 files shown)


enebo:
@dr-itz so these changes work all the way back to 5.0 but will minimize our diff for when 6.1 comes?  Also do you think we should create 60-stable yet?

dr-itz:
@enebo that deep_dup thing is an actual bug fix. Other people have seen this, also there is #954. Yes, a 60-stable would be appreciated once this thing is in and hits master.

There have been some API changes, so 60.x doesn't work anymore on Rails master.

dr-itz:
> Looks ok to me but I will ask one question: Do we care that execute_insert has changed arity? In JRuby, proper I would say this is too risky but is the Java of arjdbc an API for external consumption?

I'd say we shouldn't care too much since this is an API between the ruby and the java part...sure you can access it, but...

> We should play it safe and maintain original signature and pass context.nil for pk to the newer signature.

you lost me there. where do you mean?

dr-itz:
whoos, just notices this indented with tabs instead of spaces..will fix

enebo:
```java
@depracated
 @JRubyMethod(name = "execute_insert", required = 1)
    public IRubyObject execute_insert(final ThreadContext context, final IRubyObject sql) {
return execute_insert(context, sql, context.nil);
}
```

The only reason is if any existing external gem is using arjdbc as a transitive dep it will get an error.  I suppose the question to ask is this change so needed that we should break an existing gem (if one actually exists)?

dr-itz:
Ah now I get it. Makes a lot of sense, but we need to rename the new signature because there's the version with binds too. So old we had required = 1/2 and new we have required = 2/3 having an overlap there. Pushed on top

enebo:
@dr-itz looks fine to me now but I defer to @kares in case he realizes something I do not.

dr-itz:
@enebo thanks

dr-itz:
Something‘s going on with mariadb in CI...i‘ll check tomorrow if it‘s related

test/db/mysql/unit_test.rb
```diff
@@ -96,7 +96,7 @@ def initialize; end
       connection_handler = connection_handler_stub
 
       config = { :jndi => 'jdbc/TestDS' }
-      connection_handler.expects(:jndi_connection)
+      connection_handler.expects(:jndi_connection).with() { |c| config = c }
```
kares: just a minor thingy ... its morning here but did you not mean `config == c` ?

dr-itz: Nope. Since config is now cloned in the connection handler, we cannot check for changes on the original object. By catching the `jndi_connection` arg I can get the cloned and changed object out this way. It's a bit ugly but it was the easiest way I could come up with

kares: ooh so that is what this accomplishes, thanks!

dr-itz:
> Something‘s going on with mariadb in CI...i‘ll check tomorrow if it‘s related

ok, it's pre-existing. I'll look into it maybe over the weekend :)

PR #1041, opened 2019-11-10
[core] use RubyDateTime directly to create a string

dr-itz:
With Rails 6.1, there's no conversion to a Time anymore for a DateTime
if it's already UTC. So what happens is that we call #to_time where
we dind't before. Since Rails overrides that for DateTime to basically
do #getlocal, the result is off.

Using RubyDateTime directly avoids the problem and is faster anyway.


enebo:
@dr-itz is this valid with 6.0 as well?  We have not branch 60-stable yet (although we could from this point forward).

dr-itz:
@enebo it's valid, it just doesn't trigger with 6.0.0 yet. But the change that broke everything might get back-ported..

PR #1043, opened 2019-11-25
Rails 6.1 fixes

dr-itz:


dr-itz:
the few failing tests _should_ be fixed with #1040. I'll rebase this one once #1040 hits master

test/rails/config.yml
```diff
@@ -81,8 +81,6 @@ connections:
       username: <%= ENV['PGUSER'] || ENV['user'] || 'rails' %>
       jdbc_fetch_size: 100
       properties:
-        # Postgres' JDBC driver does not prepare statements until executed 5 times by default, let's lower:
-        prepareThreshold: <%= ENV['PREPARE_THRESHOLD'] || 1 %>
```
rdubya: Maybe this has been changed with the Rails 6 stuff, but this was originally needed to get tests to pass that tested that prepared statements were actually getting reused

dr-itz: It's not related to Rails version AFAICT. This is purely a pgjdbc thing that controls whether server-side PS are used always (when set to 1), never (when set to 0) or automatically controlled by pgjdbc (> 1, default 5). Server-side PS uses additional resources on server side and comes with an overhead. Client-side PS are still a huge advantage over using plain `Statement`s since the wire-protocol still uses bind-params and all.

Not setting `prepareThreshold` makes no difference test-wise with Rails 5.0 (except being faster). Here I clearly opt for the faster tests. Also I would argue we get better test exposure of the pgjdbc driver with the default since both the client-side and the server-side PS are tested (vs. only server-side). It doesn't even make a difference if we cache the `PreparedStatements` or not for this feature of pgjdbc to work.

That being said, it might have very well been required in the past due to a bug in earlier versions of pgjdbc itself or something.

Maybe I should leave that line in there, just make it defaults to the pgjdbc default of 5 so it's still possible to test the behavior.

dr-itz: I dropped that one for now

dr-itz:
All green now 🎉

PR #1045, opened 2019-12-03
insert_returning fails when the primary key is not the first column in postgresql

satyanash:
Add failing test for postgresql when primary key is not the first column when insert_returning.

### DIFF
--- a/test/db/postgresql/primary_key_is_not_the_first_column_test.rb
+++ b/test/db/postgresql/primary_key_is_not_the_first_column_test.rb
@@ -1,0 +1,37 @@
+require 'test_helper'
+require 'db/postgres'
+
+class PrimaryKeyIsNotTheFirstColumnTest < Test::Unit::TestCase
+
+  class CreateEmployees < ActiveRecord::Migration[4.2]
+    def self.up
+      create_table 'employees', :id => false do |t|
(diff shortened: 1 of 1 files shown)


satyanash:
Only workaround right now seems to be `insert_returning: false`. This is `true` by default when >= Rails 5.

dr-itz:
on it..basically I have a fix ready that I did for performance reasons. Need to rebase to 50-stable and test

dr-itz:
Ok, pushed out a fix + your test on top of my #1040. It's Rails 5.0, but we need that fix all the way back there. Thanks for providing the testcase!

Basically what happened is that we forgot the PG specific `sql_for_insert` stuff from AR. We never notices since the JDBC driver automatically appends `RETURNING *` to inserts. So it always just worked when the first column was the PK. Now instead of fixing up the generated SQL, instead I rely on the JDBC API.

Thinking about it, I'll move your test case to the generic part so we can test with MySQL as well.

dr-itz:
> Thinking about it, I'll move your test case to the generic part so we can test with MySQL as well.

Just checked, sqlite3 and mysql don't care, only PG is affected, so we're good.

@enebo @kares could you have a look at #1040?

satyanash:
whoa that was _fast_. Thanks!

PR #1051, opened 2020-01-25
[test] fix accessing db config hash

dr-itz:


dr-itz:
btw. the failures on 60-stable are upstream issues, fixed here: https://github.com/rails/rails/pull/38316

enebo:
Thanks for PRing to fix stuff in Rails itself!  Always good to catch early mistake commits before something gets set in stone.

PR #1057, opened 2020-03-06
Rails 6.1 fixes

dr-itz:


kares:
Nice, I think master should now reflect 6.1 dev - esp. since there's 60-stable. 
Thanks for working on these, awesome work!

PR #1058, opened 2020-03-25
[postgres] fix setting date when timezone offset is negative

dr-itz:
e.g. when running with JRUBY_OPTS=-J-Duser.timezone=America/Detroit
any date saves is a day behind. Also tested with Asia/Anadyr.


enebo:
I half wondered if RubyDate path could eliminate the "Date" check itself but then realized that was a lot of extra effort over comparing 4 characters (and retrieving them).  I realized you still need RubyDate beneath it regardless in case to_date makes a RubyDate.  So not really saying anything is wrong but thought I would share what went through my head when I was reading the change.

Issue #1065, opened 2020-05-13
Java::OrgJodaTime::IllegalInstantException on date columns


felipefava:
I recently updated `activerecord 4` to `activerecord 5` in a non Rails app with `jruby-9.2.7.0`. Now we have this version of the gem:

```
activerecord-jdbcmysql-adapter (52.4-java)
activerecord-jdbc-adapter (= 52.4)
jdbc-mysql (~> 5.1.36, < 9)
```

I tried also upgrading the version to `52.6` but we still have the issue.

The config `ActiveRecord::Base.default_timezone` value is `:utc`.

-------

The first problem we faced, was that datetimes were cast wrongly, but we fixed it with this https://github.com/jruby/activerecord-jdbc-adapter#mysql-specific-notes. More details of that issue [here](https://github.com/jruby/activerecord-jdbc-adapter/issues/1064). 

However, that fix brings us the following issue.

If we have a record with a `Date` column where the value of the date is the day where the Daylight Saving Time was made, we have this exception: 
```
Java::OrgJodaTime::IllegalInstantException (Illegal instant due to time zone offset transition (daylight savings time 'gap'): 2019-10-06T00:00:00.000 (America/Asuncion))
```
The important part of the above exception is this `2019-10-06T00:00:00.000 (America/Asuncion))`. To understand better the problem, in the timezone `America/Asuncion`, on the `2019-10-06` begun the Daylight Saving Time. And if I try to find a record with ActiveRecord that has a column date with the value `2019-10-06` it raises the mentioned exception. 

In the code of this gem, I saw that for dates columns initialize a dateTime at the beginning fo the day:
```
# RubyJbdcConnection.java -> method dateToRuby

return DateTimeUtils.newDateAsTime(context, value, null).callMethod(context, "to_date");


# DateTimeUtils.java -> method newDateAsTime

DateTime dateTime = new DateTime(year, month, day, 0, 0, 0, 0, zone);
return RubyTime.newTime(context.runtime, dateTime, 0);
``` 
I don't have experience with java and I don't know if it is necessary, but why it creates a datetime and not a date?

 I think it shouldn't fail in this case when what I want to instantiate is a date column and not a datetime.

FYI: with `activerecord 4` and `activerecord-jdbc-adapter (= 1.3.25)` this wasn't an issue

Please let me know if I'm missing something, and thanks for the help. 

------

More details of my problem:

```
jruby-9.2.7.0 :001 > Bill.find(48129508)
[...]

PR #1066, opened 2020-06-02
Fix date time zone for date columns

felipefava:
Hi! This PR is a fix for [this](https://github.com/jruby/activerecord-jdbc-adapter/issues/1065) issue


enebo:
@felipefava sorry about the delay of this and thanks for the fix!

PR #1068, opened 2020-06-14
Rails 6.1 fixes

dr-itz:
Note: one PG test still fails

### DIFF
--- a/src/java/arjdbc/jdbc/RubyJdbcConnection.java
+++ b/src/java/arjdbc/jdbc/RubyJdbcConnection.java
@@ -125,6 +125,7 @@ public class RubyJdbcConnection extends RubyObject {
     private IRubyObject config;
     private IRubyObject adapter; // the AbstractAdapter instance we belong to
     private volatile boolean connected = true;
+    private RubyClass attributeClass;
 
     private boolean lazy = false; // final once set on initialize
     private boolean jndi; // final once set on initialize
@@ -133,6 +134,7 @@ public class RubyJdbcConnection extends RubyObject {
 
     protected RubyJdbcConnection(Ruby runtime, RubyClass metaClass) {
         super(runtime, metaClass);
+        attributeClass = runtime.getModule("ActiveModel").getClass("Attribute");
     }
 
     private static final ObjectAllocator ALLOCATOR = new ObjectAllocator() {
@@ -2420,14 +2422,12 @@ public class RubyJdbcConnection extends RubyObject {
         final IRubyObject value;
         final int type;
 
-        final Integer primitiveType = jdbcTypeForPrimitiveAttribute(context, attribute);
-        if (primitiveType != null) {
-            type = primitiveType;
-            value = attribute;
-        } else {
-            //debugMessage(context, attribute);
+        if (attributeClass.isInstance(attribute)) {
             type = jdbcTypeForAttribute(context, attribute);
             value = valueForDatabase(context, attribute);
+        } else {
+            type = jdbcTypeForPrimitiveAttribute(context, attribute);
+            value = attribute;
         }
 
         // All the set methods were calling this first so save a method call in the nil case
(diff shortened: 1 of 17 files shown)


enebo:
@dr-itz There is only a single failure involving pluck and some index being the value 1 vs "1".   Is that followup work as I did not see if in the last CI run as failing?

enebo:
@rdubya @kares eef5f826391413dda238ea23c9526b580c82776d looks tasty to me but I would like a couple more eyes on that change.

dr-itz:
I looked into the pluck test. In `PostgreSQLResult.columnTypeMap` we create a map of column name => type. This works fine as long as column names are unique. But that failing has

```
Arel.sql("COALESCE(id, 0)"),
Arel.sql("COALESCE(title, 'untitled')")
```

i.e. w/o a column alias and for both column labels we get "coalesce"...second one wins.

dr-itz:
> I looked into the pluck test. In `PostgreSQLResult.columnTypeMap` we create a map of column name => type. 

looking at https://github.com/rails/rails/blob/master/activerecord/lib/active_record/result.rb#L106
it also supports an array. But that crashes when trying to load the schema. Looking into that now

dr-itz:
> looking at https://github.com/rails/rails/blob/master/activerecord/lib/active_record/result.rb#L106
> it also supports an array. But that crashes when trying to load the schema. Looking into that now

ignore me, that's for overrides only

src/java/arjdbc/jdbc/RubyJdbcConnection.java
```diff
@@ -3627,7 +3656,10 @@ protected final TableName extractTableName(
     }
 
     protected IRubyObject valueForDatabase(final ThreadContext context, final IRubyObject attribute) {
-        return attribute.callMethod(context, "value_for_database");
+        if (attribute.respondsTo("value_for_database")) {
```
kares, line 3659: would like to understand, why is this necessary - what would not respond_to? `value_for_database`?
asking since it's now double the method lookup cost and I wonder if this is a hot spot.

dr-itz:
> I haven't been keeping up with the latest ActiveRecord changes, do they support passing in primitives? Last I knew everything had to be an attribute.

Unfortunately, that's a thing now:

```ruby
        def type_casted_binds(binds)
          case binds.first
          when Array
            binds.map { |column, value| type_cast(value, column) }
          else
            binds.map do |value|
              if ActiveModel::Attribute === value
                type_cast(value.value_for_database)
              else
                type_cast(value)
              end
            end
          end
        end
```

the array case is the legacy thing we on the ruby side. 

we might well do the same check for `ActiveModel::Attribute` and then do all the thing we did until now and only use the new thing otherwise

dr-itz:
> we might well do the same check for `ActiveModel::Attribute` and then do all the thing we did until now and only use the new thing otherwise

just did that on top. Also fixed the datetime thing and a few style issues. And excluded the failing PG test for now

enebo:
Wouldn't raw values be faster than wrapping it in an attribute object?  Am I missing something?  I took this as a big possible improvement in removing an extra object boxing and indirection which would need to be inlined.

dr-itz:
Not wrapping is nice...and just for fun, I changed one line in `setStatementParameter` to always use the old `ActiveModel::Attribute` code paths. That's how a rails test run for SQLite3 looks like:

    6802 runs, 16328 assertions, 15 failures, 598 errors, 14 skips

so not wrapping seems to become a thing

src/java/arjdbc/jdbc/RubyJdbcConnection.java
dr-itz: gone with the latest version

kares:
great, another full circle from not wrapping to wrapping and now back to not wrapping ... still the "same old" AR :wilted_flower:

enebo:
@dr-itz sorry I should have merged this a week ago.  Thanks again for keeping our head above water.

PR #1076, opened 2020-11-19
Bump activerecord dependency

deivid-rodriguez:
So that Rails 6.1 prerelease can be tried against the main branch.


enebo:
Sorry @deivid-rodriguez I missed this but @dr-itz had this in the PR I just landed so it should be correct now.

deivid-rodriguez:
No problem at all, thanks @enebo!

PR #1079, opened 2020-12-14
Fix mssql failure when inserting into tables with composite primary keys

rdubya:
Noticed the test suite started failing for the activerecord-jdbcsqlserver-adapter gem after the latest version and tracked it down to MSSQL's driver not playing nice. I believe this change will need to be applied to all the branches. After we release the changes, I'll release a new version of the jdbcsqlserver gem that bumps the required version for this gem.


rdubya:
@enebo I can't remember which direction we merge, should I have pointed this somewhere higher?

enebo:
@rdubya for posterity if anyone else reads this PR.  50-stable merges to 51-stable ...  So this will hit all rails 5+ trees in the merge process.

PR #1081, opened 2020-12-16
Rails 6.1 fixes

dr-itz:


dr-itz:
@enebo would you also consider #1066?

I think after that we're ready for a first 61.0 release 🎉

enebo:
@dr-itz Fantastic.  Ok I will merge through 1066 and put out new releases since 52 and 60 will also need that one change.  61 seems important since we have had two issue reports already with our too loose 50.0 getting installed instead.  I will also make a 61-stable branch so master can be for the next version of Rails.

PR #1103, opened 2022-02-15
Bump activerecord dependency

tagliala:
So that Rails 7.0 can be tried against the main branch.

---

Hi, 👋🏼 👋🏼 

this is my first attempt to contribute to this project

While trying to upgrade ActiveAdmin to Rails 7.0, we've noticed that `activerecord-jdbc-adapter` unexpectedly downgrades to version 50.0, and that happens because it is the last version with an open dependency constraint (`activerecord >= 2.2`)

I understand that this change is probably not enough to ensure Rails 7.0 compatibility but I'm submitting because I can't find a request to add Rails 7.0 compatibility

Ref: #1076 activeadmin/activeadmin#7235


deivid-rodriguez:
This is happening at #1112, so this can be closed!

PR #1110, opened 2022-04-27
Expand GHA matrix

headius:
Expanding to other configurations from the old travis file.


headius:
With this PR we have sqlite, mysql, and pgsql running ARJDBC tests and Rails tests green.

We are still missing test runs for combinations of jruby-head, JDK 8/11/17, jndi and jdbc, different versions of PG and mysql, and Rails master, but we have a good baseline.

enebo:
@dr-itz Don't know if you noticed but we have CI running again.  Do you still have any interest in hacking on arjdbc?

dr-itz:
@enebo cool. Yes in general, it's just that I have been pretty busy otherwise, but I might find some time again soon-ish

PR #1127, opened 2023-03-06
Fix some activerecord spec failures for Rails 7 support

rdubya:
This adds support for the query tagging feature and updates the sqlite adapter to match how the typemap is created in ActiveRecord.

### DIFF
--- a/lib/arjdbc/abstract/database_statements.rb
+++ b/lib/arjdbc/abstract/database_statements.rb
@@ -10,6 +10,8 @@ module ArJdbc
       NO_BINDS = [].freeze
 
       def exec_insert(sql, name = nil, binds = NO_BINDS, pk = nil, sequence_name = nil)
+        sql = transform_query(sql)
+
         if preventing_writes?
           raise ActiveRecord::ReadOnlyError, "Write query attempted while in readonly mode: #{sql}"
         end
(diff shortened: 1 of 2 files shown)


rdubya:
It looks like a bunch of the ActiveRecord tests fail because the arguments are being recognized as a hash instead of keyword arguments. One example is:

```
TransactionalFixturesOnConnectionNotification#test_notification_established_transactions_are_rolled_back:
[1000](https://github.com/jruby/activerecord-jdbc-adapter/actions/runs/4343668791/jobs/7586001910#step:5:1001)
ArgumentError: mocked method :call expects 2 arguments, got ["book"]
```

enebo:
@rdubya The transform_query also fixed about 10-12 postgresql errors too.

The kwarg errors you mention could be a kwargs error (and they definitely are if you can see there is a kwargs hash as the extra element) but I do not see where any of that flows through our code and as far as I know our only kwargs issue would be if we have a native(Java) method which is getting called in there (we need keywords=true annotations).  I do hope that is true since that is a nice chunk of problem which wold dissolve when corrected.

Also thanks for the help.  If you have time stop by matrix/element.io channel so I can ask you about oids in postgresql :)

enebo:
@rdubya I see a failure with kwargs and mocking in minitest's unit tests.  This likely is the issue for us.

PR #1129, opened 2023-04-20
Update check_all_foreign_keys_valid for Rails 7.1

ghiculescu:
ref: https://github.com/rails/rails/pull/47990

In Rails 7.1 (when that PR is merged), there will be a new method, `check_all_foreign_keys_valid!`, which adapters can implement to check foreign keys when fixtures are loaded. This PR just implements that method. The Rails 7 `all_foreign_keys_valid?` method is also preserved.

### DIFF
--- a/lib/arjdbc/sqlite3/adapter.rb
+++ b/lib/arjdbc/sqlite3/adapter.rb
@@ -219,7 +219,22 @@ module ArJdbc
     end
 
     def all_foreign_keys_valid? # :nodoc:
-      execute("PRAGMA foreign_key_check").blank?
+      # Rails 7
+      check_all_foreign_keys_valid!
+      true
+    rescue ActiveRecord::StatementInvalid
+      false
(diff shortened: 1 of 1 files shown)


enebo:
@ghiculescu Thanks for the PR.  I just made 70-stable branch so master can be for 71 stuff.  If you have any interest in updating our GHA ci files to specify Rails 7.1 dev stuff then we can see how this change does.

ghiculescu:
I think https://github.com/jruby/activerecord-jdbc-adapter/pull/1129/commits/f784a45f40263c5282cd6144908de2a7333c9883 will to do it. But there are some red tests as is. To be honest I'm not a user of this gem, I just made this PR so we can try to avoid the breaking change in Rails 7.1.

enebo:
@ghiculescu I appreciate your help and I sort of noticed you were working through other adapters missing this.

Issue #1131, opened 2023-05-22
Updates for JDBC drivers


FraPazGal:
Hello! I wanted to check with you whether an update to the supported JDBC drivers is on the roadmap. There are several vulnerabilities affecting the used mysql and psql versions currently used that would be solved using newer ones. 

```
┌──────────────────────────────────────────────────────────────┬─────────────────────┬──────────┬───────────────────┬─────────────────────────────────┬─────────────────────────────────────────────────────────────┐
│                           Library                            │    Vulnerability    │ Severity │ Installed Version │          Fixed Version          │                            Title                            │
├──────────────────────────────────────────────────────────────┼─────────────────────┼──────────┼───────────────────┼─────────────────────────────────┼─────────────────────────────────────────────────────────────┤
│ mysql:mysql-connector-java (mysql-connector-java-8.0.27.jar) │ CVE-2022-21363      │ MEDIUM   │ 8.0.27            │ 8.0.28                          │ Difficult to exploit vulnerability allows high privileged   │
│                                                              │                     │          │                   │                                 │ attacker with network access via...                         │
│                                                              │                     │          │                   │                                 │ https://avd.aquasec.com/nvd/cve-2022-21363                  │
├──────────────────────────────────────────────────────────────┼─────────────────────┼──────────┼───────────────────┼─────────────────────────────────┼─────────────────────────────────────────────────────────────┤
│ org.postgresql:postgresql (postgresql-42.2.25.jar)           │ CVE-2022-26520      │ CRITICAL │ 42.2.25           │ 42.3.3                          │ postgresql-jdbc: Arbitrary File Write Vulnerability         │
│                                                              │                     │          │                   │                                 │ https://avd.aquasec.com/nvd/cve-2022-26520                  │
│                                                              ├─────────────────────┼──────────┤                   ├─────────────────────────────────┼─────────────────────────────────────────────────────────────┤
│                                                              │ CVE-2022-31197      │ HIGH     │                   │ 42.2.26, 42.3.7, 42.4.1         │ postgresql: SQL Injection in ResultSet.refreshRow() with    │
│                                                              │                     │          │                   │                                 │ malicious column names                                      │
│                                                              │                     │          │                   │                                 │ https://avd.aquasec.com/nvd/cve-2022-31197                  │
│                                                              ├─────────────────────┼──────────┤                   ├─────────────────────────────────┼─────────────────────────────────────────────────────────────┤
[...]

PR #1132, opened 2023-06-01
Update JDBC drivers for sqlite3, postgreql, and mysql

headius:
This updates all three drivers to latest and pulls in fixes for the CVEs listed in #1131.


kares (approved):
:+1:

PR #1137, opened 2023-09-11
Drop upper bound on jdbc-sqlite3

headius:
Having this upper bound prevents updating jdbc-sqlite3 to any recent version, which in turn prevents Rails 6.1 + JRuby + MacOS + ARM64 from working due to the older sqlite JDBC drivers not having such support.

This commit removes the upper bound so that any newer 3.x jdbc-sqlite3 can be used with Rails 6.1 + JRuby.


headius:
It's unclear to me why this was ever restricted so tightly and so specifically. The JDBC driver (for sqlite3 at least) is very stable and unlikely to have any major breaking changes.

@kares If you remember why we had this strict upper bound please let me know!

kares (approved):
fine by me, we do not rely on the driver's (specific) Java API

PR #1141, opened 2024-01-30
Update jdbc to 42.7.1

skunkworker:
Since this is a jar, this is fine to be closed for a contributor to update the `jar` (I understand for security concerns etc).


enebo:
@skunkworker I verified it is the same file (so others watching will not think I blindly did this) but you also have been visibly working through a perf issue so I didn't expect it to be another file :smile:

enebo:
@kares never quite sure on how you evaluate some of these updates and whether we go with it or skip.

kares:
@enebo used to check the CI or test locally against AR-JDBC to see if everything works before merging/releasing.
given CI has been :red_circle: for a while (all PG tests do not even start :cold_sweat:)  might be a bit harder to tell how well the driver update is supported in terms of JRuby on Rails.

enebo:
@kares yeah I should have had this rebased onto a stable branch

PR #1149, opened 2024-07-20
Update sqlite-jdbc to 3.46

segiddins:
JAR from https://github.com/xerial/sqlite-jdbc/releases/tag/3.46.0.0

### DIFF
--- a/jdbc-sqlite3/lib/jdbc/sqlite3/version.rb
+++ b/jdbc-sqlite3/lib/jdbc/sqlite3/version.rb
@@ -1,6 +1,6 @@
 module Jdbc
   module SQLite3
-    DRIVER_VERSION = '3.42.0.0'
+    DRIVER_VERSION = '3.46.0.0'
     VERSION = DRIVER_VERSION
   end
 end


segiddins:
@headius this is needed to be able to use newer SQLite features that are available in cruby from the sqlite3 gem

headius:
Great, thanks! We will merge and get a release out soon.

headius:
@enebo @kares maybe you can release this week while I'm traveling?

enebo:
I will likely find time this week.  In process of packing/cleaning so it is just dependent on that.

enebo:
Whoops.  I should have looked at the run results (which in turn is a problem with a pretty red CI) to notice this version of sqlite3 depends on slf4j-api + transitive deps.  

https://github.com/xerial/sqlite-jdbc/commit/21c77a481f6970622b99d712d36d0336d63fd554

It appears since 3.28 a lot of deps have been added (although I am not sure how many are actually needed for runtime from this page): https://central.sonatype.com/artifact/org.xerial/sqlite-jdbc/3.40.0.0/dependencies

We probably need to vendor these but at the same time I expect we will see errors due to loading this in mixed environments.

headius:
> this version of sqlite3 depends on slf4j-api + transitive deps

Ugh.

> We probably need to vendor these

If it comes to that I think we'd rather start using jar-dependencies to fetch them, so we're not shipping our own copies of these jar files.

I will also make the case to xerial/sqlite-jdbc that adding these dependencies is a major nuisance.

headius:
It looks like this concern was already raised by two others: https://github.com/xerial/sqlite-jdbc/issues/1094

I have added some fuel to that fire.

segiddins:
Given this was reverted, is there another ticket now tracking the sqlite version update?

headius:
@segiddins I have poked the issue again, but have no idea if they'll be receptive.

@enebo We may be forced to fork the project just to remove the dependency. We'd continue to merge but have our own release that just removes slf4j. I don't want to do it any more than you do!

headius:
It looks like we'll be able to get this patched in sqlite-jdbc once we get a patch to them that makes the sl4fj dependency optional. 🎉

I've asked another interested party if they could modify their slf4j removal patch to make it optional instead. If they don't do it I will.

headius:
I have pushed xerial/sqlite-jdbc#1178 to make slf4j optional. Just need to get the maintainers to merge it and release now.

anovadox:
It looks like https://github.com/xerial/sqlite-jdbc/releases/tag/3.46.1.1 has been released 🎉 
> make slf4j optional with fallback on JUL ([b8ea5ca](https://github.com/xerial/sqlite-jdbc/commits/b8ea5ca)), closes https://github.com/xerial/sqlite-jdbc/issues/1094

https://github.com/xerial/sqlite-jdbc/issues/1094#issuecomment-2372687678

headius:
When https://github.com/jruby/activerecord-jdbc-adapter/pull/1158 looks good I'll merge and release.

headius:
jdbc-sqlite-3.46.1.1 has been released!

PR #1150, opened 2024-07-25
Initial work to support Active Record 7.1 (Rails 7.1.x)

JesseChavez:
Hi @headius 

Here is some initial work to support active record 7.1.

- More than 90% of ARJDBC tests pass in my local
- The work for sqlite was already there
- The jndi connection pool callbacks is disabled temporarily,we still to work on this
- No sure if we need to update/fix `.github/workflows/ruby.yml`
- I have not run the rails tests yet. I would assume will run.

I can still look at the failing tests and rails tests later the following weeks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,7 @@ Gemfile.lock
 .settings
 activerecord-jdbc.iml
 lib/arjdbc/jdbc/adapter_java.jar
+.jrubyrc
+.rubocop.yml
+.solargraph.yml
+pik.sh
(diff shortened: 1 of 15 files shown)


enebo:
Hello @JesseChavez.  This looks great.  At one point sqlite3 was running on Rails AR tests but some change in AR itself broke that (just noting I was confused when this PR came in because I know it was running to completion -- not your problem though).  You had to commented out things which should be deleted but it is not worth holding this up for that.

PR #1152, opened 2024-07-26
Fix some arjdbc mysql tests to support AR 7.1

JesseChavez:
Hi @headius & @enebo 

Here we have more fixes.

Hopefully we will have mysql tests green.

### DIFF
--- a/lib/arjdbc/mysql/adapter.rb
+++ b/lib/arjdbc/mysql/adapter.rb
@@ -58,6 +58,11 @@ module ActiveRecord
         @connection_parameters ||= @config
       end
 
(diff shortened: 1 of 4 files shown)


enebo:
@JesseChavez cool!  

You may have noticed this but the local tests are basically a subset of tests ripped out of Rails unit tests from many many years ago.  As time moves forward tests get removed or updated for behavior.  I think there is still value with these as they are a good selection of basic behavior and simpler to green before the full AR test suite.  You can see from actual Rails test unit runs they do not catch a lot of stuff either.  Just some background.

headius:
@JesseChavez Thank you! We really appreciate the help here!

PR #1153, opened 2024-07-31
Fix some arjdbc postgres tests to support AR 7.1

JesseChavez:
Hi @headius & @enebo

Hopefully we will have postgres arjdbc tests green.

In the next coming days I will look at Rails tests


enebo:
@JesseChavez Looks like only one test now complaining about Time.zone leaking.  That is great though.  Local tests are nearly perfect now.

PR #1155, opened 2024-08-01
Rails booting up and running PostgreSQL rails tests 

JesseChavez:
Here is fixes to run postgres rails tests

- we need to use postgres 11 since rails tests are incompatible with postgres 10
- I had to skip an arjdbc failing test since is a issue in active record itself. I reproduced the issue in CRuby, https://github.com/rails/rails/issues/52485

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -84,7 +84,7 @@ jobs:
 
     services:
       postgres:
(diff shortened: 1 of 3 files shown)


enebo:
cool.  Interesting on the failure as I thought all of those were copied from an older set of AR tests.

PR #1156, opened 2024-08-06
Changes to fix many failing tests for MySQL and PostgreSQL

JesseChavez:
Hi @enebo 

Here are more changes to fix many AR tests.

Also, Rails 7.1 introduced  and new config to handle strict strings for SQLite which the jdbc driver does not support it.

I open a feature request ticket in their repo.
https://github.com/xerial/sqlite-jdbc/issues/1153

### DIFF
--- a/lib/arjdbc/abstract/database_statements.rb
+++ b/lib/arjdbc/abstract/database_statements.rb
@@ -104,7 +104,7 @@ module ArJdbc
         end
       end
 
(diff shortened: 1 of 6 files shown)


enebo:
@JesseChavez The other thing pending is another issue on sqlite-jdbc asking to remove mandatory dependency for slf4j.  That is preventing us to be at the latest version.

PR #1157, opened 2024-08-10
Relax on ActiveRecord gem version identifer.

ruyrocha:
And re-enabled test serial with trigger.

This PR is more intended to trigger the CI.


enebo:
Picked up 7.2 once relaxed.  So maybe set it back and you will see the green runs by adding the pend.

JesseChavez:
this proposed change can be closed since the issue was fix in Rails 7.1.4

seuros:
PR can be closed, we are in 8.0 now.

PR #1158, opened 2024-10-02
Update sqlite3 jdbc to 3.46.1.1

headius:
See #1149


headius:
This now looks similar to before the update. There's some test causing the suite to die early in this and the other JDBC runs, but the tests are running with the new driver. Merging.

Issue #1091, opened 2021-08-05
Error after upgrading to 61.1: ActiveRecord::JDBCError: HOUR_OF_DAY: 2 -> 3


mohamedhafez:
I'm getting the following error after upgrading from activerecord-jdbcmysql-adapter 61.0 to 61.1: `ActiveRecord::JDBCError: HOUR_OF_DAY: 2 -> 3`

It looks like its due to upgrading Connector/J from the 5.1 series to the 8.0 series. It seems like someone reported the the bug to the Connector/J bug tracker [here](https://bugs.mysql.com/bug.php?id=96276), and that it was marked as a possible duplicate of [this bug](https://bugs.mysql.com/bug.php?id=93444) that was resolved in Connector/J 8.0.22.

Currently, jdbc-mysql is using Connector/J 8.0.17, could we get an update to the latest version in the hopes of getting a fix to this bug?

Here's the stack trace I'm getting btw:

```
arjdbc/jdbc/RubyJdbcConnection.java:1082:in `execute_query'
        /var/www/vhosts/myapp/shared/bundle/jruby/2.5.0/gems/activerecord-jdbc-adapter-61.1-java/lib/arjdbc/abstract/database_statements.rb:44:in `block in exec_query'
        /var/www/vhosts/myapp/shared/bundle/jruby/2.5.0/gems/activerecord-6.1.4/lib/active_record/connection_adapters/abstract_adapter.rb:696:in `block in log'
        /var/www/vhosts/myapp/shared/bundle/jruby/2.5.0/gems/activesupport-6.1.4/lib/active_support/concurrency/load_interlock_aware_monitor.rb:26:in `block in synchronize'
        org/jruby/RubyThread.java:759:in `handle_interrupt'
        /var/www/vhosts/myapp/shared/bundle/jruby/2.5.0/gems/activesupport-6.1.4/lib/active_support/concurrency/load_interlock_aware_monitor.rb:25:in `block in synchronize'
        org/jruby/RubyThread.java:759:in `handle_interrupt'
        /var/www/vhosts/myapp/shared/bundle/jruby/2.5.0/gems/activesupport-6.1.4/lib/active_support/concurrency/load_interlock_aware_monitor.rb:21:in `synchronize'
        /var/www/vhosts/myapp/shared/bundle/jruby/2.5.0/gems/activerecord-6.1.4/lib/active_record/connection_adapters/abstract_adapter.rb:695:in `block in log'
[...]

PR #1160, opened 2024-10-22
Catch HOUR_OF_DAY error

jbaiza:
If the time stored in the missing daylight saving hour, MySQL JDBC driver starting from version 8.0.23 will raise this error (https://dev.mysql.com/doc/relnotes/connector-j/en/news-8-0-23.html).
If the date time value stored in the column actually is UTC time, then it will return a string representation of the value, not the time converted to the local time zone.
More in https://github.com/jruby/activerecord-jdbc-adapter/issues/1091

### DIFF
--- a/src/java/arjdbc/mysql/MySQLRubyJdbcConnection.java
+++ b/src/java/arjdbc/mysql/MySQLRubyJdbcConnection.java
@@ -218,6 +218,39 @@ public class MySQLRubyJdbcConnection extends RubyJdbcConnection {
     }
 
     @Override
+    protected IRubyObject timestampToRuby(final ThreadContext context,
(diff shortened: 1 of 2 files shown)


headius:
Per discussion in #1091 perhaps this is a "good enough" fix to merge now while we try to figure out a better long-term option?

jbaiza:
I will work on the new test failures to find out where the environment time zone impacted the results despite that it's set in the tests:
```
with_system_tz 'Europe/Prague' do
  Time.use_zone 'Europe/Prague' do
```

enebo:
@kares Any chance you can review this?  I think it looks ok but I don't really like seeing Exception String message comparisons (but believe me we have had to do this many times in the history of the JRuby-related projects so I am not saying it is wrong).

PR #1166, opened 2025-01-23
Some  important fixes

JesseChavez:
Hi @enebo 

The last fix is quite important since common queries would fail for example

```
find_by!(some_field: 1.4)
```

or other queries that use the statement cache.

No sure why the CI does not complete in github workflows, perhaps not enough resources or file descriptor limit, the CI completes fine in my local.

Perhaps we can branch off version 71 and release a RC version and start working in AR 7.2 in master branch.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -66,7 +66,7 @@ end
 
 group :rails do
   group :test do
-    gem 'minitest', require: nil
+    gem 'minitest', '~> 5.24.0', require: nil
     gem 'minitest-excludes', require: nil
     gem 'minitest-rg', require: nil
 
--- a/lib/arjdbc/abstract/relation_query_attribute_monkey_patch.rb
+++ b/lib/arjdbc/abstract/relation_query_attribute_monkey_patch.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require "active_model/attribute"
+
+module ActiveRecord
+  # NOTE: improved implementation for hash methods that is used to
+  # compare objects. AR and arel commonly use `[a, b] - [b]` operations and
+  # JRuby internally uses the hash method to implement that operation,
+  # on the other hand, CRuby does not use the hash method
+  # for small arrays (length <= 16).
+  class Relation
+    # monkey patch
+    module RelationQueryAttributeMonkeyPatch
+      def hash
+        # [self.class, name, value_for_database, type].hash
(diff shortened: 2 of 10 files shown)


enebo:
@JesseChavez When I browse the diff github seems to inline test run errors into and I am seeing quite a bit of this:

```text
NoMethodError: private method `hash' called for #<ActiveRecord::Relation::QueryAttribute:0x49da976>
```

I am guessing some external types are manually calling `hash` and it being marked private is generating those errors in your monkey patch?

enebo:
@JesseChavez I guess branching would not hurt since we are so far behind and these are working a lot more than they were.

JesseChavez:
@enebo 

I made the hash method public

JesseChavez:
Hi @enebo 

Do you need help with branching off ?

I am thinking if we can extract the jdbc drivers as separate gems, just an idea

like this:

https://github.com/JesseChavez/jdbc-pgsql

or 

https://github.com/JesseChavez/jdbc-mssql


I don't mean now, I mean for AR 8.0 and beyond

enebo:
@JesseChavez How would the jdbc driver differ from https://rubygems.org/gems/jdbc-postgres?

I will be making 71-stable today but I want to try and figure out what is wrong with Rails jobs on CI.  I even notices sqlite3 on one job is grabbing Rails 8+ for some reason but none of them make it to the point where we get a summary.

enebo:
ok sqlite3 figured out (not going to running commentary to a PR but since I immediately figured it out I figured I would share....it added main to the matrix and this no doubt worked until 8 where 9.4 stopped being a possibility.

JesseChavez:
@enebo Thanks for branching off now we can push fixes to support rails 7.2

`https://github.com/JesseChavez/jdbc-pgsql` is same as ` https://rubygems.org/gems/jdbc-postgres` however I have control of it.

I don't use the gems

```
activerecord-jdbcmysql-adapter
activerecord-jdbcpostgresql-adapter
activerecord-jdbcsqlite3-adapter
```

My setup is something like:

```
platforms :jruby do
  gem 'activerecord-jdbc-alt-adapter', '~> 71.0'
  gem "jdbc-mssql", "~> 12.6.0"
  gem "jdbc-pgsql", "~> 42.7.0"
  gem 'jruby-openssl'
...
```

PR #1169, opened 2025-02-06
More fixes to support Rails 7.2 

JesseChavez:
Hi @enebo 

Here is more fixes for the adapter.

I reckon the support for sqlite and Mysql is decent with exception the message pack support.

there are over 20 mysql compatibility tests failing which is not due to the adapter but to AR tests.

all of them pass with:

```
 class BaseCompatibilityTest < ActiveRecord::TestCase
   attr_reader :connection
 
+  self.use_transactional_tests = false
+
   def setup
     @connection = ActiveRecord::Base.lease_connection
     @pool = ActiveRecord::Base.connection_pool
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ adapters are available:
 
 ```yml
 development:
-  adapter: mysql2 # or mysql
+  adapter: mysql2
   database: blog_development
   username: blog
   password: 1234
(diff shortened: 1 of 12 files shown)


JesseChavez:
I reverted one commit that is causing mutex re-locking issue

look at it later

PR #1174, opened 2025-06-06
chore: Update to support Rails 8.0                                   …

seuros:
- Bump version to 80.0 for Rails 8.0.x support                                                                                                                                                                                                               - Update activerecord dependency to ~> 8.0.0                                                                                                                                                                                                              
- Update README to show master branch targets Rails 8.0.x                                                                                                                                                                                                    - Add 72-stable branch for Rails 7.2.x support                                                                                                                                                                                                            
- Require JRuby 10.0.0 minimum for Rails 8                                                                                                                                                                                                                   - Fix deprecated File.exists? to File.exist?


PS: still getting familiar with this code base.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,9 +27,10 @@ Versions are targeted at certain versions of Rails and live on their own branche
 | 61.x        | 6.1.x         | 61-stable | 9.2.7     | 8        |
 | 70.x        | 7.0.x         | 70-stable | 9.3.0     | 8        |
 | 71.x        | 7.1.x         | 71-stable | 9.4.3     | 8        |
-| 72.x        | 7.2.x         | master    | 9.4.3     | 8        |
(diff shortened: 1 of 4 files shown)


enebo:
@seuros thanks for trying to get this setup.  The PR nearly worked but it is still requiring Java 1.8 (which was what JRuby 9.4 and lower required) and JRuby 10 requires Java 21 or higher.

seuros:
Ok, will fix it in few.

seuros:
Just noticed the commit message, has the secret vim exit handshake. :D

enebo:
@seuros Thanks.  Things are not looking horrible on the test front for 8.  mysql is missing something really important but the other two seem like they may take less work to get back to green on local tests.

I am not sure if the recent zeitworks fixes to JRuby have landed but that could help some results too (older Rails test runs had some missing constant errors which are probably from that).

PR #1177, opened 2025-06-06
fix: sqlite3 fix , but omitted the time test.

seuros:
SQLite doesn't store timezone information. In fact, it doesn't even have a real datetime type.

@enebo ,  please double review the Java code. 

AR 8, expect always a Result that respond with a collection.

For the usec hack, i did not understand the test or the hack as Timezone and usec are not supported.

### DIFF
--- a/src/java/arjdbc/sqlite3/SQLite3RubyJdbcConnection.java
+++ b/src/java/arjdbc/sqlite3/SQLite3RubyJdbcConnection.java
@@ -468,6 +468,54 @@ public class SQLite3RubyJdbcConnection extends RubyJdbcConnection {
     // note: sqlite3 cext uses this same method but we do not combine all our statements
     // into a single ; delimited string but leave it as an array of statements.  This is
     // because the JDBC way of handling batches is to use addBatch().
+    // Override execute to ensure Rails 8 compatibility
+    // Rails 8 SQLite3 adapter expects execute to always return something that responds to to_a
+    @Override
+    @JRubyMethod(name = "execute", required = 1)
+    public IRubyObject execute(final ThreadContext context, final IRubyObject sql) {
+        final String query = sqlString(sql);
+        return withConnection(context, connection -> {
+            Statement statement = null;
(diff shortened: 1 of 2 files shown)


src/java/arjdbc/sqlite3/SQLite3RubyJdbcConnection.java
```diff
@@ -468,6 +468,54 @@ public IRubyObject readonly_p(final ThreadContext context) throws SQLException {
...
+            Statement statement = null;
+            try {
+                statement = createStatement(context, connection);
+
+                // SQLite3 can support multiple statements in one query
+                // Process all results but return the last one for Rails compatibility
+                boolean hasResultSet = doExecute(statement, query);
+                int updateCount = statement.getUpdateCount();
+
+                IRubyObject result = newEmptyResult(context); // Default to empty result
```
enebo, line 487: The semantics here confuse me a little bit.  If I have n statements is only the last resultset important?

If no then it seems like the result would be weird and this would be wrong.  

If yes then you are mapping the query result potentially more than once when only the last one matters.  I am guessing the point of these multiple statements is that it is likely n update/insert/delete with a query possibly at the end?

My only change (and I can land this for now) is that making potentially n newEmptyResults is extra work and if you used null you could just have an if right before `return result` make the empty result only once if null.

enebo (approved):
lol.  I put main comment into a line in this PR and not here.

src/java/arjdbc/sqlite3/SQLite3RubyJdbcConnection.java
enebo: I guess  my "if yes" paragraph implies it should only map result set for the very last statement if it only matter for the very last statement.  So a second possible change...Having said that this works so the two changes are just a way of reducing work done and we can consider them a later optimization.

enebo:
@seuros ok yeah I will take a stab at removing the extra construction.

PR #1179, opened 2025-06-06
remove invalid monkey patch require

seuros:
@enebo , switched to java 11 target .

Now we have lot of test to fix. I'm on it.


enebo:
@seuros For JRuby 10 minimum is Java 21 now so it may work with 11 but there is no reason for 11 (and if we up to 21 we can use newer Java language features in the codebase)

seuros:
I will update it,  java is versioning is confusing.

PR #1180, opened 2025-06-06
Make only last resultset process and only generate one emptyResultSet

enebo:
Try and reduce how much result set processing or ruby results are made


seuros:
Correct me if I'm wrong, but you used an early exit strategy ?

It should work.

PR #1183, opened 2025-06-06
test: removed expectations that SQLite3 understands timezones

seuros:
We could revisit this after the build is green.


seuros:
@enebo 🎉 , all ARJDBC are green.

enebo:
@seuros I am fine skipping these for now.  I know we do have 2 open Time issues on JRuby 10 issue tracker so not sure if that could be related here or not.  

This local test suite are ancient subset of AR tests that occasionally get updated as Rails changes behavior.  So it is possible these tests as written are no longer valid.

seuros:
the Timezone feature was introduced in 7.x, before that, it was just using the app timezone.

PR #1186, opened 2025-06-07
Fix version for java requirement

seuros:


headius:
It's a start! Thanks!

PR #1187, opened 2025-06-09
Update CI

seuros:
Those version are outdated and EOL.

For Mysql there is version 8.5 , but 8 is fine.

For postgresql, we use latest so we can try all the new features that don't exist in the past.


enebo:
@seuros seems reasonable to me.  If we find some issue with older postgresql that is still supported by 8.0 then we can either bump it back or add another service later.

PR #1188, opened 2025-06-11
remove anti-pattern with security issues

seuros:
This commit needs to be back-ported  in all active versions. 

7-2 7-1

### DIFF
--- a/src/java/arjdbc/jdbc/DriverWrapper.java
+++ b/src/java/arjdbc/jdbc/DriverWrapper.java
@@ -23,6 +23,7 @@
  */
 package arjdbc.jdbc;
 
+import java.lang.reflect.InvocationTargetException;
 import java.sql.Connection;
 import java.sql.Driver;
 import java.sql.SQLException;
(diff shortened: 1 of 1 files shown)


src/java/arjdbc/jdbc/RubyJdbcConnection.java
```diff
@@ -1818,10 +1818,6 @@ protected DriverWrapper newDriverWrapper(final ThreadContext context, final Stri
         catch (ClassCastException e) {
             throw wrapException(context, context.runtime.getNameError(), e);
         }
-        catch (IllegalAccessException e) { throw wrapException(context, e); }
```
enebo, line 1821: I am not sure Java exceptions will get caught higher up in Ruby here.  If so then you just crashed the Ruby process.  

@kares can you review this?

kares: Mentioned this on matrix, just adding noise to the user due nesting real cause 2 levels deeper. 

Also there aren't any security issues, it's a deprecation.

src/java/arjdbc/jdbc/DriverWrapper.java
```diff
@@ -58,9 +59,22 @@ public Driver getDriverInstance() {
         return driver;
     }
 
-    private Driver allocateDriver(final Class<? extends Driver> driverClass)
-        throws InstantiationException, IllegalAccessException {
-        return driverClass.newInstance();
+    private Driver allocateDriver(final Class<? extends Driver> driverClass) {
+        try {
+            return driverClass.getDeclaredConstructor().newInstance();
+        } catch (InvocationTargetException e) {
```
kares, line 65: could you please restore this do not do any exception handling an do it where it was before instead.

as explained already we do not want any extra `RuntimeException` or `IllegalStateException` wrapping...

PR #1193, opened 2025-06-11
fix: remove numeric deprecation

seuros:
depend on #1192

### DIFF
--- a/src/java/arjdbc/ArJdbcModule.java
+++ b/src/java/arjdbc/ArJdbcModule.java
@@ -232,7 +232,7 @@ public class ArJdbcModule {
 
         final RubyModule jdbc = runtime.getModule("Jdbc");
         if ( jdbc != null ) { // Jdbc::MySQL
-            final RubyModule constant = (RubyModule) jdbc.getConstantAt(constName);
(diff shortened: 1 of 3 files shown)


enebo (commented):
@seuros you may want to let me have a shot at these deprecations.  All of these changes are wrong but what you did was reasonable lacking context.

org.jruby.api package and @JRubyAPI annotations replaced the deprecated methods and methods changes to require ThreadContext was not documented as the replacement as this was changed in potentially thousands of methods.  For example, `getClass(String)` was made into `getClass(ThreadContext, String)`.

The obvious problem with the new API from this existing project is it pushes `Ruby` everywhere and not `ThreadContext`.  As such, this will be an extremely large PR and I might be better suited to do this.  Also, it may also inform additional API changes to JRuby itself.  I can perhaps just change one usage and then you can see the scope of changes.  

Or you can work on fixing AR failing tests for now.

seuros:
Given your expertise with JRuby internals, it makes sense for you to handle these deprecations. I was afraid that that double work is required before and after the deprecation fix, if the interface is the same. 

Then i'm 🟢 to fix the Ruby part. Closing al PRs.

PR #1195, opened 2025-06-12
fix: transaction fixes

seuros:


headius:
Why is this extra logic needed compared to the standard version?

https://github.com/rails/rails/blob/82e9029bbf63a33b69f007927979c5564a6afe9e/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb#L283-L295

headius:
I believe these changes are a subset of those merged from #1218. Let me know if you think that's not the case.

PR #1200, opened 2025-10-06
Ability to release gems for one database type at a time

mohamedhafez:
Introduces a new `DBS` env var, that will let you choose which subset of MySQL, Postgres, or SQLite you are trying to build and release gems for. Not setting it defaults to all of them, with the exception of `release:do`: for that, just to be extra safe that what is being done is intentional, I've required the caller to explicitly set it.


enebo:
@mohamedhafez Thanks for this PR.  When we get first adapter working for 8+ we can get those out quicker with this!

PR #1209, opened 2026-04-16
Fix Rails 7.2 JSONB pluck type casting

rajaravivarma-r:
Rails 7.2 started reading AR::Result column types by index during pluck, but PostgreSQLResult only stored types by column name. As a result, JSONB expressions like payload -> 'fields' exposed a named type while column_types[0] stayed nil, and pluck returned raw strings instead of cast JSON values.

Store PostgreSQL result types under both the column name and zero-based column index, and add regression coverage for JSONB array/object/scalar extraction across multiple rows in types_test.

## Behaviour examples

This change restores the behaviour we had in earlier Rails / ActiveRecord-JDBC versions, where `pluck` on JSONB SQL fragments returned type-cast Ruby values instead of raw JSON strings.

### Example 1: JSONB array extraction

```ruby
PostgresqlJsonb.pluck(Arel.sql("payload -> 'fields'"))
```

Expected behaviour in previous Rails / AR-JDBC versions, and with this change:

```ruby
[
  [{ "id" => "fld1", "name" => "Name", "type" => "singleLineText" }],
  [{ "id" => "fld2", "name" => "Title", "type" => "number" }]
]
```

Type-wise, that is:

```ruby
PostgresqlJsonb.pluck(Arel.sql("payload -> 'fields'")).map(&:class)
# => [Array, Array]
```

Behaviour seen on Rails 7.2 before this fix:

```ruby
PostgresqlJsonb.pluck(Arel.sql("payload -> 'fields'")).map(&:class)
# => [String, String]
```

### Example 2: JSONB scalar extraction

```ruby
PostgresqlJsonb.pluck(Arel.sql("payload -> 'profile' -> 'age'"))
```

Expected behaviour in previous Rails / AR-JDBC versions, and with this change:

```ruby
[25, 40]
```

Type-wise, that is:

```ruby
PostgresqlJsonb.pluck(Arel.sql("payload -> 'profile' -> 'age'")).map(&:class)
# => [Integer, Integer]
```

Behaviour seen on Rails 7.2 before this fix:

```ruby
PostgresqlJsonb.pluck(Arel.sql("payload -> 'profile' -> 'age'"))
# => ["25", "40"]
```

Internally, the issue was that AR-JDBC exposed the JSONB type under the column name, but Rails 7.2 started reading `AR::Result#column_types` by column index during `pluck`. This change stores the type under both the column name and the zero-based column index, so JSONB fragments are cast correctly again.

### DIFF
--- a/src/java/arjdbc/postgresql/PostgreSQLResult.java
+++ b/src/java/arjdbc/postgresql/PostgreSQLResult.java
@@ -98,7 +98,10 @@ public class PostgreSQLResult extends JdbcResult {
                     runtime.newFixnum(mod),
                     name);
 
-            if (!type.isNil()) types.fastASet(name, type);
+            if (!type.isNil()) {
+                types.fastASet(name, type);
+                types.op_aset(context, runtime.newFixnum(i), type);
+            }
         }
 
         return types;
--- a/test/db/postgresql/types_test.rb
+++ b/test/db/postgresql/types_test.rb
@@ -6,6 +6,7 @@ class PostgreSQLTypesTest < Test::Unit::TestCase
   OID = ActiveRecord::ConnectionAdapters::PostgreSQL::OID
 
   class PostgresqlArray < ActiveRecord::Base; end
+  class PostgresqlJsonb < ActiveRecord::Base; end
   class PostgresqlUUID < ActiveRecord::Base; end
   class PostgresqlRange < ActiveRecord::Base; end
   class PostgresqlTsvector < ActiveRecord::Base; end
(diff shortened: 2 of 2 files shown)


rajaravivarma-r:
After upgrading from Rails 7.0 to 7.2 our code started to break because of an incompatibility between Rails 7.2 and activerecord-jdbc-adapter. I investigated the problem with the help of an AI tool and made this PR with its help.

It would be nice if I could get this fixed in a activerecord-jdbc-adapter 72.x version.

Happy to modify this PR as required by the maintainer. TIA.

headius:
@skunkworker You may be in a better position to review this than me.

@rajaravivarma-r Thank you for your contribution and sorry for the delay getting this reviewed!

rajaravivarma-r:
It would be great. Thanks @headius

PR #1210, opened 2026-04-25
JRuby 10.1.0.0 compatibility

katafrakt:
Ensure compatibility with 10.1.0.0 by
- Switching RubyHash construction to factory method and don't pass the second int argument (no longer supported)
- Do the same for RubyString construction
- Add ostruct to Gemfile, as it's no longer bundled by default


headius:
Nice thanks! Had to get this done at some point and I want the 8.0 and 8.1 adapters out ASAP.

headius:
The failures appear to have been there prior to this PR so I'm going to merge.

@kares Ping me when you get a chance so we can figure out how to get CI green.

PR #1211, opened 2026-05-06
Update for postgresql driver 42.7.11

jessecai33:
- Bump postgresql driver to 42.7.11
- This should patch the vulnerability: CVE-2026-42198


headius:
Yup, good catch. I'll get updated releases of the drivers out.

headius:
jdbc-postgres-42.7.11 has been released!

PR #1215, opened 2026-06-10
Fix Rails 7.2 regression when pgbouncer backend connection is gone

### LINKED ISSUE #1213: Regression v72.1 from v71
A rails app previously on `v71.0` broke with this command when upgrading to `v72.1`

```
[ERROR] [2026-05-28 22:38:19 +0000] [Failable] ActiveRecord::JDBCError: org.postgresql.util.PSQLException: An I/O error occurred while sending to the backend.
[ERROR] [2026-05-28 22:38:19 +0000] [Failable] arjdbc/jdbc/RubyJdbcConnection.java:1100:in `execute_query'
```

I first thought this was something around the Rails 7.2 connection lease. So I introduced a `.lease_connection` at the start of each request, however the issue came back.
I found the `arjdbc.debug: true` property and have turned it on to collect more information.

skunkworker:
This is a possible patch that fixes: https://github.com/jruby/activerecord-jdbc-adapter/issues/1213

It detects the SQLState and raises a `ActiveRecord::ConnectionFailed.new` error that can be captured and handled by ActiveRecord above with the Rails 7.2 `with_raw_connection(allow_retry: true)`. 

The current ActiveRecord adapter looks for `PG::ConnectionBad` messages and looks for a newline at the end of the message (surprising).

The `pg` gem emits these [here](https://github.com/ged/ruby-pg/blob/633e16f9cde849976d6a235767423d19186a6b28/ext/pg_connection.c#L859), and tests them [here](https://github.com/ged/ruby-pg/blob/633e16f9cde849976d6a235767423d19186a6b28/spec/pg/connection_spec.rb#L369)

This patch first looks for the Postgres ErrorCodes and tries to use those first. As a fallback, some error messages are used (in situations where postgres itself is not booted, eg: a server is not online).

### DIFF
--- a/lib/arjdbc/postgresql/adapter.rb
+++ b/lib/arjdbc/postgresql/adapter.rb
@@ -737,8 +737,36 @@ module ArJdbc
       end
     end
 
+    # See https://www.postgresql.org/docs/current/errcodes-appendix.html
+    # Class 08 - Connection Exception, plus the 57P0x operator-intervention
+    # codes that signal the backend is going away.
+    CONNECTION_FAILURE_SQL_STATES = %w[
+      08000
+      08001
+      08003
+      08004
+      08006
+      08007
+      08P01
+      57P01
+      57P02
+      57P03
+    ].freeze
+    CONNECTION_FAILURE_MESSAGES = /
+      I\/O\ error\ occurred\ while\ sending\ to\ the\ backend |
+      This\ connection\ has\ been\ closed |
+      Connection\ to\ .*\ refused |
+      Connection\ is\ closed |
+      terminating\ connection\ due\ to
+    /x.freeze
+    private_constant :CONNECTION_FAILURE_SQL_STATES, :CONNECTION_FAILURE_MESSAGES
+
     def translate_exception(exception, message:, sql:, binds:)
       return super unless exception.is_a?(ActiveRecord::JDBCError)
+
+      if connection_lost?(exception)
+        return ::ActiveRecord::ConnectionFailed.new(message, sql: sql, binds: binds, connection_pool: @pool)
+      end
 
       # TODO: Can we base these on an error code of some kind?
       case exception.message
@@ -769,6 +797,18 @@ module ArJdbc
       else
         super
       end
+    end
+
+    def connection_lost?(exception)
+      state = exception.sql_state if exception.respond_to?(:sql_state)
+      return true if state && CONNECTION_FAILURE_SQL_STATES.include?(state)
+
+      message = exception.message
+      return true if message && CONNECTION_FAILURE_MESSAGES.match?(message)
+
+      cause = exception.cause if exception.respond_to?(:cause)
+      cause.is_a?(Java::JavaSql::SQLRecoverableException) ||
+        cause.is_a?(Java::JavaSql::SQLNonTransientConnectionException)
     end
 
     # @private `Utils.extract_schema_and_table` from AR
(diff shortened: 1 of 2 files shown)


lib/arjdbc/postgresql/adapter.rb
```diff
@@ -771,6 +799,18 @@ def translate_exception(exception, message:, sql:, binds:)
       end
     end
 
+    def connection_lost?(exception)
+      state = exception.sql_state if exception.respond_to?(:sql_state)
+      return true if state && CONNECTION_FAILURE_SQL_STATES.include?(state)
+
+      message = exception.message
+      return true if message && CONNECTION_FAILURE_MESSAGES.match?(message)
```
WriterZephos, line 807: Is there anything more reliable we can go on besides the message? This seems a bit brittle.

skunkworker:
ActiveRecord translates the exception from `PG::ConnectionBad`, looks for a `\n` at the end of the message (surprising to me). And if the newline is detected, it bubbles up a `ConnectionFailed` error.


https://github.com/rails/rails/blob/v7.2.3.1/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L787-L797

WriterZephos:
Wow... yeah that's less than ideal. Sounds like there isn't a better way though.

lib/arjdbc/postgresql/adapter.rb
WriterZephos: I see now that it uses state codes first. This is just a fallback. That makes a lot more sense.

lib/arjdbc/postgresql/adapter.rb
```diff
@@ -737,9 +737,37 @@ def can_perform_case_insensitive_comparison_for?(column)
...
+    private_constant :CONNECTION_FAILURE_SQL_STATES, :CONNECTION_FAILURE_MESSAGES
+
     def translate_exception(exception, message:, sql:, binds:)
       return super unless exception.is_a?(ActiveRecord::JDBCError)
 
+      if connection_lost?(exception)
+        return ::ActiveRecord::ConnectionFailed.new(message, sql: sql, binds: binds, connection_pool: @pool)
+      end
+
       # TODO: Can we base these on an error code of some kind?
```
headius, line 771: This seems ripe for a future PR to use more error codes for accuracy.

lib/arjdbc/postgresql/adapter.rb
headius, line 807: Accidentally deleted a comment from @skunkworker here:

> This is important for the scenarios where there won't be a SQLSTATE error code. Like a network failure, the remote database is down etc.
> 
> And more intended as as fallback.
> 
> * Error codes
> * Messages

lib/arjdbc/postgresql/adapter.rb
```diff
@@ -737,9 +737,37 @@ def can_perform_case_insensitive_comparison_for?(column)
       end
     end
 
+    # See https://www.postgresql.org/docs/current/errcodes-appendix.html
+    # Class 08 - Connection Exception, plus the 57P0x operator-intervention
+    # codes that signal the backend is going away.
+    CONNECTION_FAILURE_SQL_STATES = %w[
```
headius, line 743: It would be nice to generate this from something shipped by the PostgreSQL JDBC driver but I think this is ok for now.

headius (approved):
This looks clean and simple enough to me. Makes me wonder how many other message-based error checks could be replaced with sql_state codes.

headius:
@skunkworker Let me know if you have additional work planned for 72.x and we can try to schedule a release.

Issue #1173, opened 2025-04-30
Rails 8 support


osbre:
Currently, this package appears to be incompatible with Rails 8:
```bash
$ jbundle
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...
Could not find compatible versions

Because activerecord-jdbc-adapter = 71.0 could not be found in rubygems
repository https://rubygems.org/ or installed locally for any resolution
platforms (universal-java-23)
  and activerecord-jdbc-adapter >= 71.0 depends on activerecord ~> 7.1.3,
  activerecord-jdbc-adapter >= 71.0 requires activerecord ~> 7.1.3.
And because rails >= 8.0.2 depends on activerecord = 8.0.2,
  activerecord-jdbc-adapter >= 71.0 is incompatible with rails >= 8.0.2.
So, because Gemfile depends on rails ~> 8.0.2
  and Gemfile depends on activerecord-jdbc-adapter = 71.0,
  version solving has failed.

The source contains the following gems matching 'activerecord-jdbc-adapter (=
71.0)':
  * activerecord-jdbc-adapter-71.0-java
```

My `Gemfile`:

```ruby
gem 'activerecord-jdbc-adapter', '71.0', platforms: :jruby
gem 'activerecord-jdbcsqlite3-adapter', '71.0', platforms: :jruby
```

headius:
I wonder how far off the 71.x versions are from working on 7.2 and 8.0. We know we need to catch up! The jump from 7.0 to 7.1 was surprisingly complicated.

postmodern:
Just ran into this when upgrading a library to `activerecord` ~> 8.0. Also had to upgrade `sqlite3` to `~> 2.0` for ActiveRecord 8.0 under CRuby, which might imply some extra work is needed for `activerecord-jdbcsqlite3-adapter`.

pboling:
@headius I am setting this up in my extensive test suite for omniauth-identity now.  I have test against heads of Ruby Engines, like JRuby, but also heads of all required and optional runtime dependencies (like active record, sqlite3, etc).

Please make the branches for 7.2 / 8.0 so we can target them in builds, and start to see where the cracks are.

enebo:
@pboling I just made a 72-stable.  PRs for 8 can go against master now.  Thanks for pointing it out.

seuros:
@pboling: JRuby 10 needs Java 21+ now, but GitHub CI is still living in the Java 8 BC era.

[...]

Issue #1212, opened 2026-05-28
Rails 8.0+ transaction materialization breaks other tests


headius:
While working to finalize Rails 8.0 support I discovered a number of tests failing depending on the ordering of the test suite. I eventually narrowed the problems down to a set of tests for "transaction materialization" in the Rails 8.0 suite here:

https://github.com/rails/rails/blob/ff0b257d73b97d9bc5157150f93899aac3797f46/activerecord/test/cases/adapter_test.rb#L513

Several of the tests in this block use `@connection.materialize_transactions` to preserve a transaction across a reconnect. Either this logic or the associated setup/teardown of the spec causes the adapter to leave a transaction open on JRuby's adapter (tested on sqlite), usually causing many subsequent tests to fail until the system resets enough to return to a stable state.

The number of failures can reach into the thousands depending on when this block of tests runs.

If I comment out this entire block, the test suite is much more stable, and the total number of failures is in the hundreds, roughly 3-5% of all tests run.

```
8980 runs, 29717 assertions, 53 failures, 301 errors, 27 skips
```

I have seen this as low as 230-ish F/E total.

Something's clearly not right with this test block or with the `materialize_transaction` logic when run with our adapters.

headius:
My way of disabling the test:

```
diff --git a/activerecord/test/cases/adapter_test.rb b/activerecord/test/cases/adapter_test.rb
index 5bd1a06804..e9283264b6 100644
--- a/activerecord/test/cases/adapter_test.rb
+++ b/activerecord/test/cases/adapter_test.rb
@@ -510,7 +510,7 @@ def reset_fixtures(*fixture_names)
       end
   end
 
-  class AdapterConnectionTest < ActiveRecord::TestCase
+  false && class AdapterConnectionTest < ActiveRecord::TestCase
     unless in_memory_db?
       self.use_transactional_tests = false
 
```

headius:
The MySQL tests run even better than the SQLite tests with this block disabled:

```
8831 runs, 29971 assertions, 37 failures, 79 errors, 44 skips
```

ryudoawaru:
I think I've found the root cause. It's not the `materialize_transactions` logic itself — it's Rails' `raw_transaction_open?` test helper colliding with a legacy method alias on the JDBC raw connection.

**The collision**

[...]

PR #1221, opened 2026-07-18
PostgreSQL: force wire-level BEGIN when starting a transaction

ryudoawaru:
### Problem

On PostgreSQL, `begin_db_transaction` ends up calling `Connection#setAutoCommit(false)` — but in pgjdbc that only flips a client-side flag. The actual `BEGIN` is deferred and piggybacked onto the *next* statement sent (see pgjdbc's `QueryExecutorImpl#sendQueryPreamble`).

Rails 7.1+ transactions are lazy on their side too: `materialize_transactions` calls `begin_db_transaction` and expects a transaction to *actually* be open on the server afterwards. Under AR-JDBC that materialization was a wire-level no-op — the server session stays `IDLE` until the next query, so anything that inspects the raw transaction state right after materialization (e.g. Rails' own `adapter_test.rb` via `raw_transaction_open?`) sees no transaction.

### Fix

After `setAutoCommit(false)` (and setting isolation, when given), issue a no-op `SELECT 1`. pgjdbc prepends its deferred `BEGIN` to it in the same round-trip, so the cost matches the pg gem's `exec("BEGIN")` — one round-trip, just moved from the first in-transaction query to `begin_db_transaction` where AR expects it.

Includes a test that asserts the server-side transaction state is `OPEN` right after `begin_db_transaction` (it reports `IDLE` without the fix).

### Verification

Environment: JRuby 9.4.14.0, PostgreSQL 18.4, pgjdbc 42.7.11, macOS.

* `rake test_postgresql`: 306 tests, 0 failures, 0 errors (includes the new test; fails without the patch)
* Rails suite (`rails:test_postgresql`, AR `7-2-stable`), on top of `72-stable` + #1220:
  * `test/cases/adapter_test.rb`: 69 runs, **0 failures, 0 errors** — the 4 remaining failures there (`assert raw_transaction_open?` right after `materialize_transactions`, lines 533/543/553/577) are exactly what this patch fixes
  * same with `PREPARED_STATEMENTS=false`: 67 runs, 0 failures, 0 errors
  * `test/cases/transactions_test.rb`: 105 runs, 0 failures, 0 errors

Related: #1173, #1218, #1220.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/java/arjdbc/postgresql/PostgreSQLRubyJdbcConnection.java
+++ b/src/java/arjdbc/postgresql/PostgreSQLRubyJdbcConnection.java
@@ -173,6 +173,21 @@ public class PostgreSQLRubyJdbcConnection extends arjdbc.jdbc.RubyJdbcConnection
         // NOTE: only reversed order - just to ~ match how Rails does it :
         /* if ( connection.getAutoCommit() ) */ connection.setAutoCommit(false);
         if ( isolation != null ) setTransactionIsolation(context, connection, isolation);
+        // pgjdbc's setAutoCommit(false) only flips a client-side flag; the
+        // wire-level BEGIN is deferred until the next statement is sent
+        // (QueryExecutorImpl#sendQueryPreamble). Rails 7.1+ lazy transactions
+        // assume begin_db_transaction opens a server-side transaction once
+        // materialized, so flush the deferred BEGIN now with a no-op statement
+        // - pgjdbc prepends BEGIN to it in the same round-trip, matching the
+        // cost of pg's exec("BEGIN").
+        Statement statement = null;
+        try {
+            statement = connection.createStatement();
+            statement.execute("SELECT 1");
+        }
+        finally {
+            close(statement);
+        }
         return context.nil;
     }
 
--- a/test/db/postgresql/transaction_test.rb
+++ b/test/db/postgresql/transaction_test.rb
@@ -8,6 +8,20 @@ class PostgresTransactionTest < Test::Unit::TestCase
     assert_true ActiveRecord::Base.connection.supports_savepoints?
   end
 
+  # pgjdbc defers the wire-level BEGIN until the next statement is sent, but
+  # AR 7.1+ lazy transactions expect begin_db_transaction to actually open a
+  # server-side transaction (e.g. materialize_transactions), so it must flush
+  # the deferred BEGIN right away.
+  def test_begin_db_transaction_opens_transaction_on_the_server
+    connection = ActiveRecord::Base.connection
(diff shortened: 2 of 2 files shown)


headius (approved):
Looks simple enough and good to note that it has no more overhead than the equivalent `exec("BEGIN")` (I immediately had a concern about the additional round-trip).

Approved by me but would be good to get input from the other active PG contributors @skunkworker @JesseChavez.

ryudoawaru:
**A note on the failing Rails-suite checks — they are inherited from `72-stable`, not introduced here.**

The `Rails Tests (Postgres)` jobs have been red on `72-stable` for every recent push, well before this PR:

* [Jun 13 run](https://github.com/jruby/activerecord-jdbc-adapter/actions/runs/27470153275) (pre-#1218/#1219): both PS modes died mid-suite with `FATAL: sorry, too many clients already`
* [Jun 27 run](https://github.com/jruby/activerecord-jdbc-adapter/actions/runs/28289907965) (current `72-stable` HEAD, i.e. this PR's merge base): PS=true again crashed with `too many clients` before finishing; PS=false was cancelled by fail-fast

On this PR, the PS=true job actually **completes the full suite for the first time in a while** — 9481 runs, 43 failures, 74 errors ([log](https://github.com/jruby/activerecord-jdbc-adapter/actions/runs/29636832411/job/88396269651)). Those break down as:

* ~20 `ActiveRecord::AdapterConnectionTest` failures/errors: Rails' `raw_transaction_open?` test helper pokes MRI's `PG::Connection` API (`transaction_status`, `PG::PQTRANS_*`), which doesn't exist on the JDBC raw connection. #1220 adds exactly those shims — locally, `72-stable` + #1220 + this patch runs `test/cases/adapter_test.rb` at **0 failures, 0 errors** in both PS modes (69/67 runs), and `test/cases/transactions_test.rb` at 105 runs, 0F/0E. These entries should disappear once #1220 lands.
* The remaining ~90 are the pre-existing `72-stable` long tail (composite PK, hstore, explain, deferred constraints, …), unrelated to transactions.

The PS=false job died ~2 minutes in from connection-pool exhaustion, the same long-standing issue as the `too many clients` crashes above.

The `ARJDBC Tests (Postgres)` jobs (all 4 combinations) are green, including the new server-side-transaction-state test from this PR.

headius:
Thank you @ryudoawaru! I will talk with other contributors and see about getting some releases out. Do you have other fixes coming?

ryudoawaru:
Thanks for merging! Two things I'm looking at next: forward-porting this fix to master, and #1212 — I've tracked down the root cause with a reproduction and posted the analysis there. Let me know if something else has priority for the releases.

PR #1222, opened 2026-07-20
Rails 8.0 Support, Postgres fixes 

JesseChavez:
Hi @headius 

Here is the work we have done, we are looking the failures of Postgres ARJDBC own tests failures.

We will also port the below changes that are in a old MR from us, we have that changes in production and they mean to be postgres fixes only but it seems they are sqlite and MySql

```
(bash)> git log -4
commit 8fca8db0990cd2f3fbf7a161b609aafdc2ed8f80 (HEAD -> 72-stable-prod-fixes, origin/72-stable-prod-fixes)
Author: Jesse Chavez <jesse.chavez.r@gmail.com>
Date:   Sat Dec 13 11:39:18 2025 +0900

    Fix queries like where(field: [""]) to match CRuby behaviour

commit aad30c3e4be23b0a63f533f56ae536d4e5f55ee2
Author: Jesse Chavez <jesse.chavez.r@gmail.com>
Date:   Fri Oct 17 22:40:46 2025 +0900

    Comment out code setting uuid type for plain varchar field

commit cf12d83fa383daa837f7e51dc5750dc840bfa1d3
Author: Jesse Chavez <jesse.chavez.r@gmail.com>
Date:   Sun Sep 28 12:50:37 2025 +0900

    Postgres, pass type casted binds for select queries

```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -105,5 +105,6 @@ group :rails do
     gem 'minitest-rg', require: nil
     gem 'minitest-rails', require: nil
     gem 'benchmark-ips', require: nil
+    gem 'rdoc', '~> 7', require: false
   end
 end
--- a/lib/arjdbc/abstract/time_value_jruby_compat_monkey_patch.rb
+++ b/lib/arjdbc/abstract/time_value_jruby_compat_monkey_patch.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+require "active_model/type/helpers/time_value"
+
+# The caller only handles ArgumentError as a failure,
+# but JRuby raises TypeError for invalid formats in Time.new() (non-standard),
+# which isn't handled by the caller.
+# We just return nil here as it will result in fallback parsing, same as raising.
+module ActiveModel
+  module Type
+    module Helpers
+      module TimeValue
+        module JRubyCompat
+          private
+            def fast_string_to_time(string)
+              super
+            rescue TypeError
+              nil
+            end
+        end
+        prepend JRubyCompat
+      end
+    end
+  end
+end
--- a/src/java/arjdbc/jdbc/JdbcResult.java
+++ b/src/java/arjdbc/jdbc/JdbcResult.java
@@ -47,6 +47,16 @@ public class JdbcResult extends RubyObject {
         processResultSet(context, resultSet);
     }
 
+    // HACK: Needed for postgres to be able to return a sane result type instead of just a RubyFixnum
+    protected JdbcResult(ThreadContext context, RubyClass clazz, RubyJdbcConnection connection) {
+        super(context.runtime, clazz);
+
+        values = newArray(context);
+        this.connection = connection;
+        columnNames = new RubyString[0];
+        columnTypes = new int[0];
+    }
+
     /**
      * Builds a type map for creating the AR::Result, most adapters don't need it
      * @param context which thread this is running on.
--- a/src/java/arjdbc/postgresql/PostgreSQLResult.java
+++ b/src/java/arjdbc/postgresql/PostgreSQLResult.java
@@ -107,7 +107,10 @@ public class PostgreSQLResult extends JdbcResult {
                     runtime.newFixnum(mod),
                     name);
 
-            if (!type.isNil()) types.fastASet(name, type);
+            if (!type.isNil()) {
+                types.fastASet(name, type);
+                types.fastASet(runtime.newFixnum(i), type);
+            }
         }
 
         return types;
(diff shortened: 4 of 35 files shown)


headius:
Ok finally getting back to this. There's a lot here to review.

src/java/arjdbc/postgresql/PostgreSQLRubyJdbcConnection.java
```diff
@@ -324,13 +326,32 @@ protected void setArrayParameter(final ThreadContext context,
...
+            values = typeCastArrayValues(context, adapter, valueForDB);
             break;
         }
 
         statement.setArray(index, connection.createArrayOf(typeName, values));
     }
 
+    private final CachingCallSite type_cast_site = new FunctionalCachingCallSite("type_cast");
+
+    private Object[] typeCastArrayValues(final ThreadContext context, final IRubyObject adapter, final RubyArray<?> values) {
```
headius, line 341: This change concerns me; I'd like to see the equivalent code in the PG gem to be sure we're not doing excessive work here.

src/java/arjdbc/postgresql/PostgreSQLResult.java
```diff
@@ -9,13 +9,7 @@
...
 
 import arjdbc.util.PG;
-import org.jruby.Ruby;
-import org.jruby.RubyArray;
-import org.jruby.RubyClass;
-import org.jruby.RubyHash;
-import org.jruby.RubyModule;
-import org.jruby.RubyNumeric;
-import org.jruby.RubyString;
+import org.jruby.*;
```
headius: Style concern: prefer explicit imports until there's at least ten of them.

src/java/arjdbc/jdbc/JdbcResult.java
```diff
@@ -47,6 +47,16 @@ protected JdbcResult(ThreadContext context, RubyClass clazz, RubyJdbcConnection
         processResultSet(context, resultSet);
     }
 
+    // HACK: Needed for postgres to be able to return a sane result type instead of just a RubyFixnum
```
headius, line 50: Need to figure out something better than a HACK for this. Note also concerns in the same commit about this not being the best impl. Let's figure out the best impl.

lib/arjdbc/abstract/time_value_jruby_compat_monkey_patch.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require "active_model/type/helpers/time_value"
+
+# The caller only handles ArgumentError as a failure,
+# but JRuby raises TypeError for invalid formats in Time.new() (non-standard),
```
headius, line 6: I'd like to have a JRuby bug reference here so we can revert this patch when JRuby gets fixed.

test/rails/excludes/mysql2/ActiveRecord/ConnectionAdapters/PoolConfig/ResolverTest.rb
```diff
@@ -0,0 +1 @@
+exclude :test_url_invalid_adapter, "sqlserver is not a built-in adapter, so rails complains when validating the error"
```
headius, line 1: Very annoying that the Rails tests implicitly require there be no other adapters. This warrants a patch to Rails imho.

test/rails/excludes/postgresql/ActiveRecord/ConnectionAdapters/ReaperTest.rb
```diff
@@ -1,2 +1,3 @@
 exclude :test_some_time, 'intermittent failures, leaks thread, fires at high frequency'
 exclude :test_connection_pool_starts_reaper, 'intermittent failures, leaks thread, fires at high frequency'
+exclude :test_reaper_works_after_pool_discard, 'deadlocks under JDBC due to high frequency discard'
```
headius, line 3: Do we know why this deadlocks? Of course we have lots of excludes but this seems like one we should try to fix rather than excluding.

Gemfile
```diff
@@ -105,5 +105,6 @@ group :rails do
     gem 'minitest-rg', require: nil
     gem 'minitest-rails', require: nil
     gem 'benchmark-ips', require: nil
+    gem 'rdoc', '~> 7', require: false
```
headius, line 108: This should be temporary; we need to work with rbs folks to get a release out with JRuby support.

headius:
Also a general note: it's hard to evaluate this with all the 8.0 jobs failing.

